PR #165, opened 2014-10-27
Support various request methods for /delay

tvbuehler:
Add `POST`, `PUT`, `DELETE`, `PATCH` and `TRACE` methods to the allowed list for `/delay/<int:delay>`; also return `json` (similar to the special methods for the added request types).


tvbuehler:
No other handler accepts these; I think I copied the list from `'/status/<codes>'`. Not sure about `OPTIONS`, but handling `HEAD` doesn't make much sense imho: the request body would be the interesting part.

PR #197, opened 2014-12-23
new digest-auth-nocookie route and improved robustness of digest-auth

larkost:
This adds a new route (`/digest-auth-nocookie`) that skips the cookie written at the 401 authentication requests. Additionally it fixes username checking, avoids some cases that were causing crashes (500 responses), and does a better job of reporting missing cookies as well as crashes (500 responses).

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -370,31 +370,31 @@ def hidden_basic_auth(user='user', passwd='passwd'):
 
 
 @app.route('/digest-auth/<qop>/<user>/<passwd>')
-@app.route('/digest-auth-nocookie/<qop>/<user>/<passwd>', defaults={'checkCookie':False})
-def digest_auth(qop=None, user='user', passwd='passwd', checkCookie=True):
+@app.route('/digest-auth-nocookie/<qop>/<user>/<passwd>', defaults={'check_cookie':False})
+def digest_auth(qop=None, user='user', passwd='passwd', check_cookie=True):
     """Prompts the user for authorization using HTTP Digest auth"""
     if qop not in ('auth', 'auth-int'):
         qop = None
     try:
-        remoteAddr = request.remote_addr or u''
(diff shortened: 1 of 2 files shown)


httpbin/core.py
```diff
@@ -370,34 +370,50 @@ def hidden_basic_auth(user='user', passwd='passwd'):
...
+            if check_cookie is True:
+                response.headers['Set-Cookie'] = 'auth=%s' % remote_addr
+            return response
+        elif check_cookie is True and request.cookies.get('auth') != remote_addr:
+            # check for auth challange cookie per https://github.com/Runscope/httpbin/issues/124
+            response = app.make_response('Missing the cookie set in the 401 response. '
+                'This client seems broken. To bypass this check use the digest-auth-nocookie route.')
+            response.status_code = 403
+            return response
+    except Exception as e:
```
larkost: Before this change you got nothing back, simply an empty 500 exception. I haven't toured the other handlers much, but this one does a lot more calculations than I would expect elsewhere, and those were not created with a mind to make sure that the input was correct. I fixed some holes, but I would have to re-write everything from scratch to be certain of understanding it well enough to cover all of the bases. So instead I made it easier to diagnose the problem when it came up again.

sigmavirus24: I'd rather the code base not use anything like `except Exception`. If more problems come up, we can figure it out by running it locally in debug mode after we get a bug report. Please remove this.

reinert:
I really would like to see this feature merged.
I will definitely use it for my client showcase (both with cookie and without).
What is missing for it to be accepted? Only the query param issue?

httpbin/core.py
larkost: Without it the exception goes out as a 500 error with no explanation. This still gives the 500 error, but adds more data so that someone using it can get the information. Why would you want to toss information that you might not be able to reproduce locally?

PR #199, opened 2014-12-24
Add `httpbin` console_script

msabramo:
Makes it easier to run in a simple (non-productiony) way to get a feel for what it does:

```
$ httpbin
 * Running on http://127.0.0.1:5000/
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,11 @@ setup(
     author='Runscope',
     author_email='httpbin@runscope.com',
 
+    entry_points = """\
+      [console_scripts]
+      httpbin = httpbin.core:app.run
+    """,
+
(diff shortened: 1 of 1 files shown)


sigmavirus24:
> httpbin is much easier and cleaner and future-proof than python -m httpbin.core

We can make `python -m httpin` work instead. I'm not sure where you get `future-proof` from though.

msabramo:
It's cool with me. FWIW, I don't know what I was thinking with future-proof. Guess I was imagining a hellish world with no python -m.

sigmavirus24:
There's a hellish past where that's the case, but I doubt it's going to disappear.

PR #202, opened 2015-01-16
Add range request functionality

dshirley:
A basic implementation of range requests:
http://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7233.html

Allows querying http://httpbin.org/range-request/1024?duration=s&chunk_size=code, specifying Range headers that allow selecting specific portions of the resource to GET.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -21,7 +21,7 @@ from werkzeug.wrappers import BaseResponse
 from six.moves import range as xrange
 
 from . import filters
-from .helpers import get_headers, status_code, get_dict, parse_request_range, check_basic_auth, check_digest_auth, secure_cookie, H, ROBOT_TXT, ANGRY_ASCII
+from .helpers import get_headers, status_code, get_dict, get_request_range, check_basic_auth, check_digest_auth, secure_cookie, H, ROBOT_TXT, ANGRY_ASCII
 from .utils import weighted_choice
 from .structures import CaseInsensitiveDict
 
@@ -564,21 +564,7 @@ def range_request(numbytes):
     pause_per_byte = duration / numbytes
 
     request_headers = get_headers()
-    request_range = parse_request_range(request_headers['range'], numbytes)
-
-    first_byte_pos = request_range[0]
-    last_byte_pos = request_range[1]
-    if first_byte_pos is None and last_byte_pos is None:
-        # Request full range
-        first_byte_pos = 0
-        last_byte_pos = numbytes - 1
-    elif first_byte_pos is None:
-        # Request the last X bytes
-        first_byte_pos = max(0, numbytes - last_byte_pos)
-        last_byte_pos = numbytes - 1
-    elif last_byte_pos is None:
-        # Request the last X bytes
-        last_byte_pos = numbytes - 1
+    first_byte_pos, last_byte_pos = get_request_range(request_headers, numbytes)
 
     if first_byte_pos > last_byte_pos or first_byte_pos not in xrange(0, numbytes) or last_byte_pos not in xrange(0, numbytes):
         response = Response(headers={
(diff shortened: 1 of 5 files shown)


httpbin/core.py
```diff
@@ -537,10 +537,91 @@ def generate_bytes():
             yield(bytes(chunks))
 
     headers = {'Transfer-Encoding': 'chunked',
-               'Content-Type': 'application/octet-stream'}
+               'Content-Type': 'application/octet-stream' }
```
sigmavirus24: Why did you change this? It's not a style improvement and it has no other benefits. Please remove the trailing space.

dshirley: ah, I think this was an accident.

sigmavirus24:
Why does this need to be a separate endpoint? You're duplicating portions of other endpoints (it seems). Why not abstract that code into a separate module and use those common portions instead of copying code around?

dshirley:
Yep, you're correct.  This endpoint is an amalgamation of code in /stream-bytes and /drip.
I would much prefer that this functionality IS a part of one (or both) of these endpoints.  However, I didn't do that because it requires changing the behavior of those endpoints.  The behavior changes might be negligible enough to where they don't matter - I'll let you be the judge.

Here's the requirements that didn't jive with the other end points.

1)  The payload generation has to be deterministic.  /stream-bytes generates random data.  /stream-bytes even allows for a random seed.  The seed parameter would have to be removed.
2)  I need chunked transfer encoding.  This doesn't seem to be in the RFC, but one of the things I'm hoping to test is NSURLSession downloadTaskWithResumeData (https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSURLSession_class/#//apple_ref/occ/instm/NSURLSession/downloadTaskWithResumeData:).  It seems that Apple requires the transfer-encoding be "chunked" in order for this to work.  The bottom line is "/drip" currently uses "identity" encoding rather than chunked.
3) The data has to be a predictable length.  /stream-bytes takes the ceiling of the input number of bytes with 100k.  It would be pretty wonky to request bytes 100k-200k and also have the endpoint automatically clip the amount of data you requested.

There is one other benefit and that is having a separate endpoint makes it a bit more obvious which endpoint supports range-requests - much like the "stream-bytes" endpoint makes it obvious which endpoint supports chunked transfer-encoding.

Are you guys ok with modifying the behavior of any of the other endpoints?  I'm also fine factoring out the generating function, though it is fairly different than the other generating functions, so I'm not sure how much benefit will come of it as far as code deduplication goes (though it maybe more readable).

Thoughts?

johnsheehan:
i'll let you all sort out the refactoring issues...my only question here is the path name. is `-request` required or would just `/range` suffice?

dshirley:
@johnsheehan I changed the endpoint from /range-request to /range

dshirley:
@sigmavirus24 Do you have any particular suggestions on what should be factored out?  The only piece that is really copy/pasted is the part that gets the chunk_size from the URL parameters:

      if 'chunk_size' in params:
          chunk_size = max(1, int(params['chunk_size']))
      else:
          chunk_size = 10 * 1024

There are a few other places where I borrowed inspiration, but the logic seems different enough that refactoring the code would require some unnatural gymnastics that wouldn't necessarily result in more readable code.

I did factor out a piece of logic to make the code a bit more concise.  I'm open to further changes.  I just want to make sure we're in sync with the direction we want it to go before I spend time doing the work.

dshirley:
@johnsheehan Can this be merged?

johnsheehan:
@sigmavirus24 and others, any last comments on the code? i'm cool with the functionality and route name.

sigmavirus24:
@johnsheehan one request and then I'm okay with this.

dshirley:
@johnsheehan @sigmavirus24 I removed the changes to the .gitignore by doing a "git reset HEAD~1; git push -f".

sigmavirus24:
Last I checked I was :+1: on the code

johnsheehan:
will deploy to httpbin.org in the next week or two (swamped with other projects now)

dshirley:
Good working with you guys.  Much appreciated!

PR #206, opened 2015-01-23
Expose WWW-Authenticate header for CORS & Add hidden-digest-auth service

### LINKED ISSUE #205: Digest authentication doesn't work for CORS
Digest auth endpoint is missing the "Access-Control-Expose-Headers: WWW-Authenticate" header in order to correctly support CORS requests. Without it, the browser doesn't allow the client to get the value of the WWW-Authenticate header.

### LINKED ISSUE #209: Create a 'hidden-digest-auth'
Add a new service "hidden-digest-auth" similar to "hidden-basic-auth" returning 404 instead of 401. This is specially necessary because digest authentication responding with 401 in ajax applications will cause the browser to always prompt the user for the credentials in the first attempt.

reinert:
Added the header Access-Control-Expose-Headers=WWW-Authenticate to the unauthorized response of digest authentication to enable ajax applications to have access to this header and perform the digest auth. Fixes #205.

Also created the 'hidden-digest-auth' service which returns 404 status coded instead of 401, preventing the browser to prompt the user for the credentials in ajax applications. Closes #209.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -434,6 +434,7 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
         auth.set_digest('me@kennethreitz.com', nonce, opaque=opaque,
                         qop=('auth', 'auth-int') if qop is None else (qop, ))
         response.headers['WWW-Authenticate'] = auth.to_header()
+        response.headers['Access-Control-Expose-Headers'] = 'WWW-Authenticate'
         response.headers['Set-Cookie'] = 'fake=fake_value'
         return response
     return jsonify(authenticated=True, user=user)
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -74,6 +74,7 @@ if os.environ.get("BUGSNAG_API_KEY") is not None:
 def set_cors_headers(response):
     response.headers['Access-Control-Allow-Origin'] = request.headers.get('Origin', '*')
     response.headers['Access-Control-Allow-Credentials'] = 'true'
+    response.headers['Access-Control-Expose-Headers'] = 'WWW-Authenticate'
 
     if request.method == 'OPTIONS':
         # Both of these headers are only used for the "preflight request"
@@ -411,7 +412,7 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
     """Prompts the user for authorization using HTTP Digest auth"""
     if qop not in ('auth', 'auth-int'):
         qop = None
-    if 'Authorization' not in request.headers or  \
+    if 'Authorization' not in request.headers or \
                        not check_digest_auth(user, passwd) or \
                        not 'Cookie' in request.headers:
         response = app.make_response('')
@@ -434,7 +435,6 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
         auth.set_digest('me@kennethreitz.com', nonce, opaque=opaque,
                         qop=('auth', 'auth-int') if qop is None else (qop, ))
         response.headers['WWW-Authenticate'] = auth.to_header()
-        response.headers['Access-Control-Expose-Headers'] = 'WWW-Authenticate'
         response.headers['Set-Cookie'] = 'fake=fake_value'
         return response
     return jsonify(authenticated=True, user=user)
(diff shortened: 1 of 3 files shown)


sigmavirus24:
+0. I'd rather there tests with this.

reinert:
Could you please review this PR?
I'm using in my local env successfully, but I need it published in httpbin.org so I can refer it in my [client showcase](http://reinert.io/requestor/latest/examples/showcase#authentication) (I'm already consuming many httpbin services).

sigmavirus24:
Looks good to me. Anyone else from @Runscope/httpbin want to review this before it's merged?

kevin1024:
:+1:

johnsheehan:
I'm good with it. I think we should add this to every endpoint where it doesn't interfere with other functionality. Is there a good way to do this more broadly (or is it already?)?

kevin1024:
> Is there a good way to do this more broadly

Sure, could add it here: https://github.com/Runscope/httpbin/blob/master/httpbin/core.py#L76

reinert:
I don't think it would be a good idea (conceptually) because exposing specific headers is too specific to the service itself. I don't see any other service that need to expose the 'WWW-Authenticate' header. Yet, others might need to expose different headers, what would cause some confusion - "where is the 'Access-Control-Expose-Headers' header being set? At service level, or filter level?"

reinert:
Do you all agree to move the 'Access-Control-Expose-Headers' to a filter level? If so, I can change it. Otherwise, what is missing to the PR be accepted? (I will make other two for the cookie and hidden issues)

johnsheehan:
I'm only suggesting moving the CORS header to be more universal. The Digest
auth specific headers should only be on this endpoint.

reinert:
I've added the hidden digest service requested in #209. (I forgot to mention it in the commit message. Sorry about that).

sigmavirus24:
So we'll need to update this Pull Request description before it's merged to reflect the reality of the scope.

Please no one merge this until @reinert rectifies the description with the reality.

sigmavirus24:
Also @reinert it looks like you rebased and in doing so re-committed a165278. I won't merge this and have to copies of the same commit in the history. Please fix your rebased PR.

reinert:
I didn't knew you merge the commits from github, sorry about that. (I'm used to rebase locally and push to github)

I've rebased my tree as requested. 
Title has been updated. The same for initial description.
Anything else?

httpbin/core.py
```diff
@@ -409,35 +410,22 @@ def hidden_basic_auth(user='user', passwd='passwd'):
...
-        ]))
-        opaque = H(os.urandom(10))
-
-        auth = WWWAuthenticate("digest")
-        auth.set_digest('me@kennethreitz.com', nonce, opaque=opaque,
-                        qop=('auth', 'auth-int') if qop is None else (qop, ))
-        response.headers['WWW-Authenticate'] = auth.to_header()
-        response.headers['Set-Cookie'] = 'fake=fake_value'
-        return response
+        return digest_challenge_response(qop, 401)
```
sigmavirus24: This needs to wait for the relevant part of #197 to land.

reinert: Why not invert the process? I could implement the optional cookie feature if desired too.

sigmavirus24: @reinert I'd rather you not re-implement something that's already been implemented by someone else and is functionally their contribution. Please don't do that.

reinert: That's right. Yet, I don't see one contribution blocking another. The one
that is good to merge can be accepted and the other can adjust itself to
it. This adjustment must happen from one of the two parts anyway. Be it by
the last finished. Otherwise, if the first never finishes, the second would
never be accepted although it's already finished.
Em sáb, 7 de fev de 2015 às 13:36, Ian Cordasco <notifications@github.com>
escreveu:

reinert:
@sigmavirus24 #197 hasn't been updated since 23 Dec 2014. How long will you wait until it is resolved? I'm just waiting this PR in order to make a release. Thanks.

reinert:
Any updates?

reinert:
@sigmavirus24 @kevin1024 @johnsheehan please give us some feedback. It has been already two months with no updates. Please consider moving on the PR pipeline.

heat:
:+1:  I'm using yours service and need this feature as soon as you guys can add.

PR #208, opened 2015-01-26
Allow httpbin to recognize SSL connections from more reverse proxy headers

### LINKED ISSUE #207: Incorrect URI scheme reported
As of some point within the last week the HTTPS version httpbin has started to report an incorrect URL with HTTPS swapped for HTTP, in its response JSON. 
At a guess this the result of some reverse proxying? (use X-Forwarded-Proto?)

See example below, where ```https://httpbin.org/post``` becomes  ```http://httpbin.org/post```.

```
$ curl -v -X POST "https://httpbin.org/post"
* Adding handle: conn: 0x6f7078
* Adding handle: send: 0
* Adding handle: recv: 0
* Curl_addHandleToPipeline: length: 1
* - Conn 0 (0x6f7078) send_pipe: 1, recv_pipe: 0
* About to connect() to httpbin.org port 443 (#0)
*   Trying 54.175.222.246...
* Connected to httpbin.org (54.175.222.246) port 443 (#0)
* successfully set certificate verify locations:
*   CAfile: C:\Users\zool1112\AppData\Local\Atlassian\SourceTree\git_local\bin\curl-ca-bundle.crt
  CApath: none
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server key exchange (12):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using DHE-RSA-AES256-SHA
* Server certificate:
*        subject: OU=Domain Control Validated; OU=EssentialSSL Wildcard; CN=*.httpbin.org
*        start date: 2015-01-16 00:00:00 GMT
*        expire date: 2016-01-16 23:59:59 GMT
*        subjectAltName: httpbin.org matched
*        issuer: C=GB; ST=Greater Manchester; L=Salford; O=COMODO CA Limited; CN=COMODO RSA Domain Validation Secure Server CA
*        SSL certificate verify ok.
> POST /post HTTP/1.1
> User-Agent: curl/7.30.0
> Host: httpbin.org
> Accept: */*
>
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Server: nginx
< Date: Mon, 26 Jan 2015 10:30:03 GMT
< Content-Type: application/json
< Content-Length: 280
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
<
{
  "args": {},
  "data": "",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.30.0",
    "X-Forwarded-Ssl": "on"
  },
  "json": null,
  "origin": "129.67.24.143",
  "url": "http://httpbin.org/post"
}
* Connection #0 to host httpbin.org left intact
```

RyanGWU82:
There's no standard for the headers added by a reverse proxy when it handles
an https incoming request. Some proxies add `X-Forwarded-Proto` but others add
`X-Forwarded-Protocol` or `X-Forwarded-Ssl`. This change lets httpbin work with
reverse proxies use any of these headers. It also instructs httpbin to ignore
those headers because they're added by a proxy.

Fixes #207. @sigmavirus24 was correct, Heroku uses `X-Forwarded-Proto` but the
Runscope hosting platform uses `X-Forwarded-Protocol`.

### DIFF
--- a/httpbin/helpers.py
+++ b/httpbin/helpers.py
@@ -39,6 +39,8 @@ ENV_HEADERS = (
     'X-Heroku-Queue-Depth',
     'X-Real-Ip',
     'X-Forwarded-Proto',
+    'X-Forwarded-Protocol',
+    'X-Forwarded-Ssl',
     'X-Heroku-Queue-Wait-Time',
     'X-Forwarded-For',
     'X-Heroku-Dynos-In-Use',
@@ -139,12 +141,16 @@ def semiflatten(multi):
 def get_url(request):
     """
     Since we might be hosted behind a proxy, we need to check the
-    X-Forwarded-Proto header to find out what protocol was used to access us.
+    X-Forwarded-Proto, X-Forwarded-Protocol, or X-Forwarded-SSL headers
+    to find out what protocol was used to access us.
     """
-    if 'X-Forwarded-Proto' not in request.headers:
+    protocol = request.headers.get('X-Forwarded-Proto') or request.headers.get('X-Forwarded-Protocol')
+    if protocol is None and request.headers.get('X-Forwarded-Ssl') == 'on':
+        protocol = 'https'
+    if protocol is None:
         return request.url
     url = list(urlparse(request.url))
-    url[0] = request.headers.get('X-Forwarded-Proto')
+    url[0] = protocol
     return urlunparse(url)
 
 


sigmavirus24:
:+1:

darrelmiller:
@RyanGWU82 [RFC 7239](https://tools.ietf.org/html/rfc7239) was approved last year as the official way of doing this.  It's likely going to be a while before it gets widespread adoption, but at least there is a standard now.

PR #212, opened 2015-02-05
IP check

failedguidedog:
Added an endpoint that helps one validate Origin IP against specified *valid* or *invalid* IP address or network.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,7 @@ Endpoint                                 Description
 .. _/get: http://httpbin.org/get
 .. _/: http://httpbin.org/
 .. _/ip: http://httpbin.org/ip
-.. _/ip/check: http://httpbin.org/ip/check?valid=127.0.0.1/32
+.. _/ip/check?valid=ip&invalid=ip: http://httpbin.org/ip/check?valid=127.0.0.1/32
 .. _/gzip: http://httpbin.org/gzip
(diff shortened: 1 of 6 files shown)


sigmavirus24:
The tests are failing so there's a problem there. Further, I'm not quite certain this makes sense in HTTPbin. Until there's further discussion, I'm strongly :-1:

failedguidedog:
@sigmavirus24 I'm aware of the failing tests :( For some reason, 'py34' tests fail because 'ipaddr' can't be installed (https://travis-ci.org/Runscope/httpbin/jobs/49579278). Strangely, the tests pass locally when they are run together with 'py27' environment.

Regarding if this makes sense in context of HTTPbin: As far as I'm concerned, this is a feature I needed and it made sense to implement it in this context. I'm contributing it back, because maybe somebody else will need it too. Specifically, this is the only way to implement IP checks when using HAProxy.

I'm willing to put in some more time to figure out why tests fail, but beyond that, I'm happy with using my own fork.

johnsheehan:
Appreciate the contribution! I don't think it's something we want to put on httpbin.org, so I don't think we should merge it in.

sigmavirus24:
So I think cases like this would be easier to handle if we had a clear documentation about the scope of the project. We should sit down and figure out what we want httpbin to be moving forward and how we want to maintain it.

PR #218, opened 2015-03-02
Add method to support additional cookie breadcrumb values

robinske:
Allows testing for additional breadcrumbs: `max_age`, `expires`, `domain`, etc.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -375,6 +375,29 @@ def set_cookies():
     return r
 
 
+@app.route('/cookies/set', methods=['PUT'])
+def set_advanced_cookies():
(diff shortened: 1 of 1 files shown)


httpbin/core.py
```diff
@@ -375,6 +375,29 @@ def set_cookies():
     return r
 
 
+@app.route('/cookies/set', methods=['PUT'])
```
sigmavirus24: Probably a stupid question on my part, but why choose PUT over POST?

robinske: Should allow overwriting of cookies on the response (for cookies with the same "key")

PR #222, opened 2015-04-05
Fix MANIFEST.in for README and templates/images

ant1441:
* Fix README filename
* Use recursive-include for templates dir to make sure images are included


johnsheehan:
thanks!

Issue #224, opened 2015-04-23
Allow POST to /delay


tmenier:
The /delay endpoint currently returns a 405 (method not allowed) on POST. Any chance you could open it up to allow it? Thanks.

sigmavirus24:
Can you enlighten us as to why this is necessary?

johnsheehan:
We've done this for our version running at yourapihere.com. I'm personally
a fan of opening up any of the non-method endpoints to allow any method. I
just haven't gotten around to it yet.

The use case is that many POST requests take awhile to process data and
this would be able to simulate that.

+1 for the request if you want to do a pull req

sigmavirus24:
@johnsheehan sure, I guess I'm just trying to understand what is trying to be tested. Every HTTP library I've used across a couple different languages doesn't care if there's a delay in using a specific method. So, since HTTPBin is usually used to test things, I'm trying to understand the behaviour being tested.

tmenier:
Thanks guys. @sigmavirus24 I agree that the method shouldn't matter in a delay/timeout situation, I'm just trying to prove it to settle a [bug report](https://github.com/tmenier/Flurl/issues/32).

Thanks for this service, it's been extremely valuable in testing my project. Glad to contribute a pr, thought I'd need to learn me a little python but it looks pretty clear.

sigmavirus24:
@tmenier yeah it makes sense to me. FWIW, if you're going to send the PR you'll be changing [this line](https://github.com/Runscope/httpbin/blob/master/httpbin/core.py#L442) to look a bit more like [this other line](https://github.com/Runscope/httpbin/blob/master/httpbin/core.py#L288). You will need to specifically mention `'GET'` in your paramter to `methods` though, otherwise you'll break the existing API.

Cheers,

tmenier:
Thanks for the tip. For completeness there's probably several other endpoints where this change might make sense for the same reasons, but I only did /delay for this PR.

PR #228, opened 2015-05-09
Allow any HTTP method in /delay endpoint

tmenier:
Issue #224

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -439,7 +439,7 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
     return jsonify(authenticated=True, user=user)
 
 
-@app.route('/delay/<int:delay>')
+@app.route('/delay/<int:delay>', methods=['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'TRACE'])
 def delay_response(delay):
     """Returns a delayed response"""
     delay = min(delay, 10)
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -439,7 +439,7 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
     return jsonify(authenticated=True, user=user)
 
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
LGTM. @kevin1024 @johnsheehan thoughts?

johnsheehan:
what about `HEAD` and `OPTIONS`?

i think we should create a base list of methods and expand this to apply to all the non-method specific endpoints (basically everything except `/get`, `/post`, etc)

tmenier:
I'd defer to you guys for implementation (or someone who knows python better than me), but to do this en masse it might make sense to make "any method" the default and decorate only those you want to limit (/get, /post, etc).

darrelmiller:
@tmenier  I suggest we drop the HEAD method because HEAD is [not allowed to return a body](https://tools.ietf.org/html/rfc7231#section-4.3.2).  Some HTTP client libraries get upset when a HEAD returns a body and end up returning a 504.

sigmavirus24:
I concur regarding dropping HEAD as an accepted method. That said, I'm curious about 

> Some HTTP client libraries get upset when a HEAD returns a body and end up returning a 504.

Which libraries? I'm disappointed (not surprised) that libraries would modify the status code and misrepresent the response to the user.

darrelmiller:
@sigmavirus24 It's not that they changed the status arbitrarily.  They are just failing and there is no standard way to say "this server isn't serving valid HTTP".  This URL http://conference.hypermediaapi.com tries to return a body with HEAD.  I just tried with four different HTTP request tools.  One returns 500, the other 504, and the two chrome extensions say "no response".

tmenier:
Sorry for the delay (no pun intended). This sort of fell off my radar. HEAD removed as suggested.

fflorent:
Any news? That would be very handy

Florent

kennethreitz:
feel free to re-open with a rebase!

PR #230, opened 2015-05-20
Wrote Dockerfile (Python)

kennethreitz:


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV HOME /app
+ENV PORT 3000
+ENV PATH /app/.heroku/python/bin:/tmp/python-pack/bin:$PATH
+ENV STACK cedar-14
+ENV PYTHONHOME /app/.heroku/python
+ENV PYTHONPATH /app/
+ENV DOCKER_BUILD 1
+
+
+RUN mkdir -p /app/.heroku
```
kevin1024: I recommend merging these together into a big chain (ie `mkdir -p /app/.heroku && mkdir -p /tmp/app && mkdir -p /app/src && etc` in order to avoid creating a large number of intermediate images when building.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN mkdir -p /app/.profile.d
+RUN mkdir -p /tmp/python-pack
+RUN mkdir -p /tmp/cache
+RUN mkdir -p /tmp/environment
+
+
+WORKDIR /app/src
+WORKDIR /tmp/python-buildpack/bin
+
+WORKDIR /app/
```
kevin1024: Looks like some duplicate WORKDIRs here.  No reason to have them like this since they only change all RUN,CMD,ENTRYPOINT,COPY,AND ADD statements that follow.

Dockerfile
kennethreitz: Great idea!

Dockerfile
kennethreitz: fantastic!

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ONBUILD COPY . /app/
```
kevin1024: I don't think you need to use `ONBUILD` for these unless there is another image you are creating that will derive from this.  For example, the python images have `ONBUILD` instructions that install `requirements.txt`  so that when you use `FROM python` your image will automatically install it without explicitly running the `pip install -r requirements.txt` command.

kevin1024:
I'm glad to see this; I think being able to run in Docker easily will be a nice addition.

kennethreitz:
@kevin1024 this was created with `heroku docker:init`. You can suggest these fantastic changes over here: 

https://github.com/heroku/heroku-docker

sigmavirus24:
Didn't we already reject a Dockerfile from @msabramo?

kevin1024:
Looks like it's still open #216

PR #233, opened 2015-05-27
Added POST request with delayd response

umer-master:
I have added a new endpoint so that we can test a delayed POST request response.


johnsheehan:
This doesn't warrant a new endpoint. We're adding POST support to `/delay` instead. https://github.com/Runscope/httpbin/issues/224

umer-master:
@johnsheehan Thank you. I did not know there was some work already going on regarding this

PR #238, opened 2015-07-08
make link_page return 404 if offset is to big

nkoehring:
I don't know if it is wanted, but I in my oppinion it makes way more sense, if /links/5/10 returns a 404. Then one is able to test his link grabber (or whatever) to only work on existing links for example.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -631,6 +631,9 @@ def link_page(n, offset):
     """Generate a page containing n links to other pages which do the same."""
     n = min(max(1, n), 200) # limit to between 1 and 200 links
 
+    if (offset >= n):
+        return status_code(404)
+
     link = "<a href='/links/{0}/{1}'>{2}</a> "
 
     html = ['<html><head><title>Links</title></head><body>']


sigmavirus24:
So this apparently needs a rebase. I think it makes sense, but I'll let other @Runscope/httpbin cores weigh in before merging it.

johnsheehan:
Failed range validation usually falls under 400 Bad Request. Would that work?

We should also update the docs to explain what the last path segment does.

sigmavirus24:
> Failed range validation usually falls under 400 Bad Request. Would that work?

That was my first instinct. The thing is the first number is the number of links to place on the page (in HTML), e.g., `/links/20` will put 20 links (hrefs) on the page.

The second number is the link for each link on the page, e.g., `<a href="/links/20/0">0</a> <a href="/links/20/1">1</a> ...<a href="/links/20/19">19</a>`. So while range validation kind of makes sense and a 400 bad request makes sense, a 404 kind of makes sense too. `/links/20/[0:20]` exist but `/links/20/[20:]` don't. Since they don't strictly exist they can be 404 or 400.

nkoehring:
What @sigmavirus24 said. I decided for 404 out of the "user" point of view. I personally would prefer that because you would use it for testing something which would check for static sites and therefor exepct a 404.

johnsheehan:
OK I'll buy that. Will merge once docs are in place

sigmavirus24:
> Will merge once docs are in place

Also once it's been rebased ;)

PR #239, opened 2015-07-10
Use url_for to resolve path instead of hard coding it

tatsuhiro-t:
Using url_for is useful if httpbin is served under a prefix path.  For
example, we can serve httpbin under http://example.org/httpbin.
Previously, we have to replace all hard coded paths, which produces
many diffs, and it makes hard to upgrade httpbin from source.  This
commit uses url_for to resolve correct path even if prefix path is
used.  For digest auth, we have to prepend request.script_root to
request.path to pass authentication.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -249,10 +249,10 @@ def relative_redirect_n_times(n):
     response.status_code = 302
 
     if n == 1:
-        response.headers['Location'] = '/get'
+        response.headers['Location'] = url_for('view_get')
         return response
 
-    response.headers['Location'] = '/relative-redirect/{0}'.format(n - 1)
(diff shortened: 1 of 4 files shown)


sigmavirus24:
I'm on board with this up until the change to httpbin.1.html in the templates. I'm not sure that renders the way you expect it to but I'll pull this down tonight and test it locally to make sure it does.

Thanks @tatsuhiro-t!

tatsuhiro-t:
Thank you for quick response.  I tested both settings: running httpbin directly and with prefix path "httpbin" using DispatcherMiddleware, and both worked fine, but yeah, there is plenty of space I'm missing something.

sigmavirus24:
Looks good to me. Thanks @tatsuhiro-t

PR #242, opened 2015-08-12
Set 10MB limit on /drip

mansilladev:


kevin1024:
Should we document this limit somewhere?  Maybe in the readme?

mansilladev:
@kevin1024 - good call. Added to readme.

sigmavirus24:
Sorry, could you provide some reasoning behind this? I'm not opposed I just want to make sure I understand the motivation appropriately given there's bug report this apparently closes or description of the pull request in the original comment.

mansilladev:
Large byte size requests time out after just several MB of xfer in most cases. Also, not having a cap could possibly attract DoS or overflow attack.

sigmavirus24:
Thanks @mansilladev! 🍰 Sorry for the delay

PR #246, opened 2015-08-31
Turn off tracking unless HTTPBIN_TRACKING is set.

### LINKED ISSUE #245: Tracking scripts in httpbin package on PyPI
It seems the package released to PyPI contains tracking beacons for Google Analytics and perfectaudience.com

I can see that being useful for httpbin.org (even though I'd prefer another service than Google Analytics, but whatever) - but I really think the version released to PyPI (which could get used e.g. for running it via a CI for some kind of tests) should come without those.

The-Compiler:
Fixes #245.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -93,7 +93,7 @@ def set_cors_headers(response):
 def view_landing_page():
     """Generates Landing Page."""
 
-    return render_template('index.html')
+    return render_template('index.html', env=os.environ)
 
 
 @app.route('/html')
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -92,8 +92,8 @@ def set_cors_headers(response):
 @app.route('/')
 def view_landing_page():
     """Generates Landing Page."""
-
-    return render_template('index.html', env=os.environ)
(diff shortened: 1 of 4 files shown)


httpbin/core.py
```diff
@@ -93,7 +93,7 @@ def set_cors_headers(response):
 def view_landing_page():
     """Generates Landing Page."""
 
-    return render_template('index.html')
+    return render_template('index.html', env=os.environ)
```
johnsheehan: Instead of passing the whole shebang, how about just a `tracking_enabled` option? Will keep the template cleaner too

The-Compiler:
I moved the environment variable handling out of the template as requested - this now turns tracking on if `HTTPBIN_TRACKING` is set to anything in the environment, but I think that makes more sense than pretending that's a boolean.

The-Compiler:
Any update on this, @johnsheehan @kevin1024? :smile:

sigmavirus24:
Thanks @The-Compiler!

I'd release this but I don't have access to do so.

kevin1024:
OK, new version is up on pypi, thanks @The-Compiler 

@sigmavirus24 I added you to pypi so you can do releases in the future.

Issue #236, opened 2015-06-01
/image returns 404 if accept header does not include valid media type


darrelmiller:
This should either return 406 - Not Acceptable or return a default media type.

sigmavirus24:
Could also return 400 - Bad Request

darrelmiller:
@sigmavirus24  My poor choice of words.  I should have said "does not include a supported media type".  I'm not suggesting the media type supplied is badly formatted.  Anyway, this is not a big issue.  Probably not worth changing the behavior without some specific reason to.

sigmavirus24:
Woops, I should have grokked that the first time around. I'm actually very much in favor of making httpbin do the right thing and follow mostly best practices (where they make sense for the project). I'm :+1: on a 406 being returned here personally.

phouse512:
Is anyone working on this? I can take a stab at this if nobody is.

phouse512:
See PR #250 @darrelmiller @sigmavirus24

PR #250, opened 2015-09-15
Added 406 unsupported media type response to /image

phouse512:
Based on the conversation from Issue #236 - also includes a basic message to let the user know what happened.

### DIFF
--- a/httpbin/helpers.py
+++ b/httpbin/helpers.py
@@ -54,6 +54,14 @@ ROBOT_TXT = """User-agent: *
 Disallow: /deny
 """
 
+ACCEPTED_MEDIA_TYPES = [
+    'image/webp',
+    'image/svg+xml',
+    'image/jpeg',
+    'image/png',
+    'image/*'
+]
+
 ANGRY_ASCII ="""
           .-''''''-.
         .' _      _ '.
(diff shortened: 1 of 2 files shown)


sigmavirus24:
So, this is good. I'm trying to find if there is a way to communicate to the user what media types we deem acceptable. It's looking like there is no header specifically for this. Perhaps we should just throw it in the body of the response?

phouse512:
Yea I was looking [here] (http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) at the 406 error code and it says the following:

>Unless it was a HEAD request, the response SHOULD include an entity containing a list of available entity characteristics and location(s) from which the user or user agent can choose the one most appropriate.

Is the 406 only thrown by for bad image types? If so I can just add a list of the supported ones to the data body.

sigmavirus24:
Please stop referencing 2616. 2616 is dead. [This](https://tools.ietf.org/html/rfc7231#section-6.5.6) is what you want.

phouse512:
@sigmavirus24 woops my mistake, I added the body content looks like this: ![Image of terminal] (https://static.dyp.im/F4jIw34s9r/ed1109f08c9bc7af83838ddc9715281e.png)

How does that look?

sigmavirus24:
Is there a reason not to return something better like JSON? Something structured lilke:

```
{
  "message": "Client did not request a supported media type.",
  "accept": [
    "image/webp",
    "image/svg+xml",
    "image/jpeg",
    "image/png",
    "image/*"
  ]
}
```

phouse512:
Good call, it now returns it in that format as json.

httpbin/helpers.py
```diff
@@ -209,6 +217,10 @@ def status_code(code):
                 'x-more-info': 'http://vimeo.com/22053820'
             }
         ),
+        406: dict(data=json.dumps({
```
sigmavirus24: 2 last nits:

1. Can you update the message to be closer to a sentence like in my example
1. Can you also add the appropriate headers, e.g.,

   ```py
   dict(data=json.dumps({
             # ...
         }),
         headers={'Content-Type': 'application/json'})
   ```

phouse512:
:+1: there we go!

sigmavirus24:
Thanks @phouse512

PR #253, opened 2015-10-04
Add ability to specify port/host to run on

s-chizhik:
I guess, it would be good, if users will have ability to run `httpbin` on custom port/host.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -189,7 +189,7 @@ Or run it directly:
 
 .. code:: bash
 
-    $ python -m httpbin.core
+    $ python -m httpbin.core [--port=PORT] [--host=HOST]
 
 Changelog
 ---------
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Let's use argparse instead. I don't think we should try to divine the meaning of extra parameters to the start-up script.

sigmavirus24:
I'm okay with this. I'd like to know if @darrelmiller can chime in on whether or not the defaults here will work with @Runscope/httpbin 's deployment mechanisms.

darrelmiller:
@sigmavirus24 I wish I could tell you, but I'm not at Runscope anymore.  Maybe @mansilladev  can help.

sigmavirus24:
Sorry @darrelmiller. Didn't notice that. Congrats on the new gig!

s-chizhik:
@sigmavirus24 ping? :)

sigmavirus24:
No one from Runscope chimed in so I guess they don't care.

PR #254, opened 2015-10-05
Add a Brotli endpoint.

Lukasa:
This pull request adds an endpoint that supports the [Brotli](http://google-opensource.blogspot.co.uk/2015/09/introducing-brotli-new-compression.html) compression algorithm used as a content-encoding, alongside the current support for gzip and deflate. Chromium have announced their [intention to implement](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/xdVm8c2GOMQ) Brotli, and Firefox 44 included [their support for Brotli](https://bugzilla.mozilla.org/show_bug.cgi?id=366559). Curl also [intend to implement Brotli](http://curl.haxx.se/docs/todo.html#More_compressions). I have opened this pull request myself because the urllib3 project intends to support Brotli as well (see shazow/urllib3#713), and we needed an endpoint to test against: httpbin seemed like a good choice.

Google claims that Brotli achieves 20%-26% higher compression ratios than other compression algorithms. They claim the reference implementation decodes just as quickly as deflate in zlib, but with denser compression than LZMA and bzip2.

This implementation currently relies on the hyper project's [brotlipy](https://github.com/python-hyper/brotlipy) sub-project, a CFFI-based wrapper around the Brotli reference implementation. This is the most likely reason you may not be interested in this endpoint: it adds an extra third-party dependency, and even trickier that dependency is a CFFI-based wrapper. Brotlipy currently ships with wheels for most Windows/OS X systems, but _will_ require an install-time compile on Linux (sorry, there really isn't much we can do about that right now).

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -204,6 +204,15 @@ def view_deflate_encoded_content():
         'origin', 'headers', method=request.method, deflated=True))
 
 
+@app.route('/brotli')
+@filters.brotli
+def view_brotli_encoded_content():
+    """Returns Brotli-Encoded Data."""
+
+    return jsonify(get_dict(
+        'origin', 'headers', method=request.method, brotli=True))
+
+
 @app.route('/redirect/<int:n>')
 def redirect_n_times(n):
     """302 Redirects n times."""
--- a/httpbin/filters.py
+++ b/httpbin/filters.py
@@ -10,6 +10,8 @@ This module provides response filter decorators.
 import gzip as gzip2
 import zlib
 
+import brotli as _brotli
+
 from six import BytesIO
 from decimal import Decimal
 from time import time as now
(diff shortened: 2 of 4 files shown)


Lukasa:
Oh, BTW: the PyPy tests will fail because Brotlipy depends on a CFFI version greater than 1.0.0, and Travis ships a PyPy that's too old to use it. There's nothing that can be done about this until Travis update their PyPy version.

sigmavirus24:
I'm +1 on including this. I think there's a way to trick Travis into using a more recent PyPy with pyenv, but I don't know it off the top of my head. Perhaps we can add that in a separate PR and then recheck this once that's merged?

kennethreitz:
rebase?

Lukasa:
Rebased.

kennethreitz:
✨🍰✨

PR #259, opened 2015-10-22
Do not forcefully emit Transfer-Encoding header

Lukasa:
Discovered when attempting to use pytest-httpbin with requests (see kennethreitz/requests#2184 and kevin1024/pytest-httpbin#27).

It turns out that the WSGI specification *specifically forbids* implementations from emitting the `Transfer-Encoding` header, as noted in [this part of PEP 333](https://www.python.org/dev/peps/pep-0333/#other-http-features) (emphasis in the original):

> WSGI applications **must not** generate any "hop-by-hop" headers, attempt to use HTTP features that would require them to generate such headers, or rely on the content of any incoming "hop-by-hop" headers in the `environ` dictionary.

Now, it's extremely clear that certain WSGI servers allow the use of generators to generate chunked transfer encoding, and gunicorn is one of them. More importantly, they do this *regardless* of whether the application attempts to set the `Transfer-Encoding` header or not: this is simply how gunicorn does chunked transfer encoding. However, on WSGI servers that *don't* do this (e.g. wsgiref), attempting to emit a `Transfer-Encoding` header can cause a 500 error.

For this reason, httpbin simply shouldn't insist on emitting the header itself: instead, let the web server do it.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -280,7 +280,6 @@ def stream_n_messages(n):
             yield json.dumps(response) + '\n'
 
     return Response(generate_stream(), headers={
-        "Transfer-Encoding": "chunked",
         "Content-Type": "application/json",
         })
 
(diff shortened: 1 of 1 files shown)


kevin1024:
:+1: from me.  Since this is blocking kennethreitz/requests#2184 and causes a bug in pytest-httpbin I'm happy to do a merge+release, as long as there are no objections from @Runscope/httpbin.

johnsheehan:
I trust you all on this one. Once it's merged I'll deploy it.

johnsheehan:
Deployed along with all of this 

image

PR #261, opened 2015-11-04
Added support for another Digest algorithm, SHA256.

Jaaap:
RFC 7616 specifies other algorithms for Http Digest:
http://tools.ietf.org/html/rfc7616

I've added support for multiple algorithms in the URLs:
Old:
https://httpbin.org/digest-auth/auth/user/passwd
New:
https://httpbin.org/digest-auth/MD5/auth/user/passwd
https://httpbin.org/digest-auth/SHA-256/auth/user/passwd

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -405,9 +405,11 @@ def hidden_basic_auth(user='user', passwd='passwd'):
     return jsonify(authenticated=True, user=user)
 
 
-@app.route('/digest-auth/<qop>/<user>/<passwd>')
-def digest_auth(qop=None, user='user', passwd='passwd'):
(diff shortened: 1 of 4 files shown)


sigmavirus24:
I'm all for supporting new algorithms, but breaking HTTPbin for existing users is entirely unacceptable. We need a new way to do this.

johnsheehan:
Yep, open to the functionality but can't break existing URLs. Maybe a URL
param?

Jaaap:
I've committed a fix (82e67ed) which changes the URL's to

/digest-auth/auth/user/passwd/MD5
and also
/digest-auth/auth/user/passwd
still works.


Teun

sigmavirus24:
Thanks @Jaaap!

PR #266, opened 2015-11-25
Floating point type conversion

luhkevin:
Apologies, but my last pull request, which was merged in https://github.com/Runscope/httpbin/commit/4763d4761407dce3a99be3c6640c14dcd4be30fe, was merged with a bug.

Since the int constraint is removed to allow for floats, the delay parameter comes in as unicode. 
I'm not exactly sure how the unicode is compared with the integer 10, but this obviously creates undesirable delays -- 10 ends up being smaller than whatever number the min function interprets the unicode as.

I've run this with delay between 1 - 10 and 0 - 1 and they all return 10.

Sorry again for the trouble. Hopefully there won't be any more.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -441,7 +441,7 @@ def digest_auth(qop=None, user='user', passwd='passwd'):
 @app.route('/delay/<delay>')
 def delay_response(delay):
     """Returns a delayed response"""
-    delay = min(delay, 10)
+    delay = min(float(delay), 10)
 
     time.sleep(delay)
 


sigmavirus24:
Can you please remove the extra commits from this? The only one that needs to be here is de877ac. If you need instructions on how to do that, let me know.

luhkevin:
Actually yeah, how do you exactly get rid of the extra commits? 

I was fiddling around with the pull request interface and the branch you merged but wasn't able to get it down to just that one commit.

sigmavirus24:
So take a look at the output of the following:

```
$ git remote -v
```

It might look like

```
origin ... git@github.com:luhkevin/httpbin.git ...
origin ... git@github.com:luhkevin/httpbin.git
```

So you'll want to then do `git remote add upstream https://github.com/Runscope/httpbin` and then `git remote update`. You'll then see something like 


```
origin ... git@github.com:luhkevin/httpbin.git ...
origin ... git@github.com:luhkevin/httpbin.git ...
upstream ... https://github.com/Runscope/httpbin ...
upstream ... https://github.com/Runscope/httpbin ...
```

Now what you want to do is `git checkout master` (since you seem to be using a `dev` branch) and `git pull upstream master`. Then do `git checkout -b pull/266` and `git cherry-pick dev` (which will grab the last commit on dev). Then you can switch back to `dev` and do `git reset --hard pull/266` and finally `git push --force origin HEAD`.

In general, you always want to checkout master and do `git pull upstream master` and then create a new branch each time from master. This will make pull requests cleaner and keep you up-to-date with everything happening on the upstream project.

luhkevin:
Neat, thanks for all the help! I will follow those guidelines the next time I make a pull request.

sigmavirus24:
Thanks @luhkevin for fixing up your PR and this bug. =D

johnsheehan:
Can you bump the version (0.4.1 is good) and update the changelog too?
Similar to
https://github.com/Runscope/httpbin/commit/c688307868a2ace65b29b1e6d43b593ca154c913
Separate
PR is fine. Once that's merged, I'll deploy to httpbin.org

johnsheehan:
Deployed

luhkevin:
Cool!

PR #268, opened 2015-12-23
Upgrade gevent to 1.0.2

GarrettHeel:
We recently updated the version of python we use internally to run httpbin and have upgraded gevent for this bug fix:
https://github.com/gevent/gevent/issues/477


sigmavirus24:
:shipit:

PR #280, opened 2016-03-04
Add basic /short endpoint.

fragmede:
Reports a content-length of 2 but only returns a single byte.

Useful for testing a bad/lossy connection or a server that returns malformed data.


kennethreitz:
`/too-short` :)

sigmavirus24:
I think we're encroaching on pathod's territory here.

kennethreitz:
ALL THE ENDPOINTS!

sigmavirus24:
Yeah, this is the kind of thing pathod is excellent for. Thanks for sending the pull request though @fragmede but I'm afraid we can't accept this.

PR #281, opened 2016-03-23
Add a new Request /longdelay/n

DerPate:
Hey folks just want to give back a Addition we make locally to test even longer timeouts with our Android App. You can close the Pull Request when you don't like the longer delays


sigmavirus24:
I doubt Runscope would want to deploy this to their production instance given it could be a DDoS vector.

That said, I wonder if we should make delay configurable from the environment so that this kind of local patching is unnecessary.

DerPate:
Thx, for at least your Response but whats the different in an DDoS Vector Threat between 6 10s delays compared to one 60s delay just to get my head around

PR #282, opened 2016-04-06
QEA-136: Added endpoint to create/get file on a server

suchitparikh:
- POST: creates a file of `name` is `name` argument is passed. If not, creates a file
with unique name
- GET: Returns the file content file `name` if `name` argument is passed.  If no `name`
is passed, it returns the content of all the files in multipart/form-data response.
Returns up to a `limit` number of files
Returns files updated since last `created_since` seconds

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,8 @@
 httpbin(1): HTTP Request & Response Service
 ===========================================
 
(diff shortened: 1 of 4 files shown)


suchitparikh:
PR created by mistake against origin

PR #288, opened 2016-05-06
Do not save and check cookies for Digest authentication. 

VavrikVladimir:
The cookies does not relate to Digest.
I'm working on software which does not handle cookies in http client, so the digest didn't work for me.


sigmavirus24:
We can't just remove features from httpbin. Other HTTP client's and software rely on that check to assert the correctness of their client.

PR #293, opened 2016-06-30
add license title

waldyrious:
It's not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)


sigmavirus24:
Thanks!

PR #294, opened 2016-06-30
added other methods to the delay endpoint, added tests for delay

stdunbar:
A pull request to add other HTTP methods to the "delay" endpoint.  For testing it is nice to be able to test a slow down for more than just a GET.  I added POST and PUT.  Perhaps others could be added but I know that these are useful.  I also added some tests for the three delay methods.

I notice that the unit tests are failing but it doesn't appear to be in the code I changed.  test_response_headers_multi and test_response_headers_simple are both failing.


stdunbar:
Sorry - didn't see the other PR until I created this.

Issue #307, opened 2016-10-19
Error 500 on digest-auth with qop='auth-int'


felixpalta:
This is **NOT** a duplicate of #162 

**Steps to reproduce the problem:**
- run httpbin locally

`gunicorn httpbin:app`
- Attempt to access digest authorization test, using qop='auth-int'

`curl http://localhost:8000/digest-auth/auth-int/user/pass --digest -u user:pass  -v`

**Result:**

> GET /digest-auth/auth-int/user/pass HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8000
> Accept: _/_
> 
> < HTTP/1.1 401 UNAUTHORIZED
> < Server: gunicorn/19.6.0
> < Date: Wed, 19 Oct 2016 12:05:41 GMT
> < Connection: close
> < Content-Type: text/html; charset=utf-8
> < Content-Length: 0
> < WWW-Authenticate: Digest nonce="b93e7a83b26a02875c89f2dda4a7169e", realm="me@kennethreitz.com", algorithm=MD5, opaque="b6d3c9f2db6b642bfd451c1ca32a5bcb", qop="auth-int"
> < Set-Cookie: fake=fake_value
> < Access-Control-Allow-Origin: *
> < Access-Control-Allow-Credentials: true
> < 
> 
> GET /digest-auth/auth-int/user/pass HTTP/1.1
> Authorization: Digest username="user", realm="me@kennethreitz.com", nonce="b93e7a83b26a02875c89f2dda4a7169e", uri="/digest-auth/auth-int/user/pass", cnonce="Y2NlMjRhNDNjNjc0ODAwMzAwMGMxNGM5MDAwNTJjNTU=", nc=00000001, qop=auth-int, response="73d0d9ad422d57e72a2820d6d3f85c78", opaque="b6d3c9f2db6b642bfd451c1ca32a5bcb", algorithm="MD5"
> User-Agent: curl/7.35.0
> Host: localhost:8000
> Accept: _/_
> 
> < HTTP/1.1 500 INTERNAL SERVER ERROR
> < Server: gunicorn/19.6.0
> < Date: Wed, 19 Oct 2016 12:05:41 GMT
> < Connection: close
> < Content-Type: text/html
> < Content-Length: 291
> < 
> <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
> <title>500 Internal Server Error</title>
> 
> <h1>Internal Server Error</h1>
[...]

PR #308, opened 2016-10-19
GH-307 Fix error 500 on digest auth with qop='auth-int'

felixpalta:
Fix typo in HA2() function, which caused Digest auth test to fail, when 'qop' parameter is set to 'auth-int'.

### DIFF
--- a/httpbin/helpers.py
+++ b/httpbin/helpers.py
@@ -292,9 +292,10 @@ def HA2(credentails, request, algorithm):
         for k in 'method', 'uri', 'body':
             if k not in request:
                 raise ValueError("%s required" % k)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Please add tests for this. Also, folks do not need the module name (or any other non-standard shorthand) in the commit message. Please simplify your message to "Pass algorithm parameter to hash function" with an explanation of *why* in the paragraph of the message.

felixpalta:
@sigmavirus24 I have added extensive tests for all combinations of Digest Authentication (DA) parameters, also tested manually with curl.

sigmavirus24:
@felixpalta thanks so very much! That's awesome!

PR #311, opened 2016-11-29
Add endpoint /echo

povilasb:
`/echo` returns request body.
E.g.
```
$ curl -v http://localhost:5000/echo -d "body"

> POST /echo HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:5000
> Accept: */*
> Content-Length: 4
> Content-Type: application/x-www-form-urlencoded
> 
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 4
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
< Server: Werkzeug/0.9.4 Python/3.5.2
< Date: Tue, 29 Nov 2016 20:11:31 GMT
< 
body 
```

Such endpoint becomes useful when we are testing proxy servers - how they are handling request body, etc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,6 +57,7 @@ Endpoint                                 Description
 `/forms/post`_                           HTML form that submits to */post*
 `/xml`_                                  Returns some XML
 `/encoding/utf8`_                        Returns page containing UTF-8 data.
+`/echo`_                                 Echoes request body.
 ======================================   ==================================================================================================================
 
 .. _/user-agent: http://httpbin.org/user-agent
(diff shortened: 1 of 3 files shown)


sigmavirus24:
The `/post` endpoint returns the `body` to you as it came to it (just inside some JSON with more data like headers, etc. that came on the request). So what benefits does this have over that endpoint?

povilasb:
You're fast :)

I use it for automated testing.
Such endpoint makes testing simpler  - I don't need to parse json and extract body field.

johnsheehan:
Use mocky.io, mockable.io, jsonstub.com or another response mocking tool for this instead of httpbin.

PR #314, opened 2016-12-24
Return content-length in 'range' requests

keflavich:
For 'range' requests, knowing the original content-length is important for some tests I'm running.  So I've added content-length to the returned headers, where that length is set to the value requested in the URL.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -598,7 +598,8 @@ def range_request(numbytes):
         response = Response(headers={
             'ETag' : 'range%d' % numbytes,
             'Accept-Ranges' : 'bytes',
-            'Content-Range' : 'bytes */%d' % numbytes
+            'Content-Range' : 'bytes */%d' % numbytes,
+            "Content-Length": str(numbytes),
             })
         response.status_code = 416
         return response
@@ -625,7 +626,9 @@ def range_request(numbytes):
         'Content-Type': 'application/octet-stream',
         'ETag' : 'range%d' % numbytes,
         'Accept-Ranges' : 'bytes',
-        'Content-Range' : content_range }
+        "Content-Length": str(numbytes),
+        'Content-Range' : content_range
+    }
 
     response = Response(generate_bytes(), headers=response_headers)
 


kennethreitz:
✨🍰✨

nateprewitt:
@kennethreitz, given the description of this endpoint on httpbin:

>Streams n bytes, and allows specifying a Range header to select a subset of the data. Accepts a chunk_size and request duration parameter.

This seem fundamentally incorrect for the /range endpoint. It may be worth considering reverting this. `Content-Length` shouldn't be present on a streamed response.

keflavich:
Is the problem just that `chunk_size` is being ignored?  I don't see any reason `content-length` shouldn't be returned for data with a known size, though I understand why it should be absent for streamed data with no known length.

nateprewitt:
So the problem is that the `/range` endpoint is specifically defined to *stream* bytes at a given chunk size over n seconds. This is accomplished by using the `chunked` transfer encoding, so the client can read periodically, keeping discrete chunks rather than combining everything into a single blob. This change has circumvented the `206` functionality, and really the point of having the delayed trickle. You receive a single dump from the server because of the nature of `Content-Length` which makes the `/range` endpoint equivalent to `/bytes` with a delay parameter.

I'd suggest we add a delay param to `/bytes` rather than breaking `/range` if that's the need functionality.

keflavich:
OK, I certainly see the need for having both a streaming (no length) and a non-streaming (has length) endpoint that accepts the `bytes` header keyword.  I'm not sure I know how to implement that.

nateprewitt:
Hey @keflavich, before we make any changes lets step back for a moment.

Could you describe your test cases and what kind of 'range' requests you're trying to perform that need the content length? Is the delay important to you here, or are you just looking for responses with `Content-Range`/`Accept-Ranges` headers? It's just striking me as odd to use `/range` if you're requesting all bytes every time.  Example requests and expected responses would be super helpful.

keflavich:
@nateprewitt I meant to add that information with the previous post.  I'm attempting to test a module that resumes partially completed requests.  It needs to check that the file has not been completely downloaded first.  So, in pseudocode:

```
session = requests.Session()
result = session.get(url, stream=True)
existing_file_length = os.stat(local_filepath).st_size
if existing_file_length >= length:
    # all done!
else:
    session.headers['Range'] = "bytes={0}-{1}".format(existing_file_length,
                                                  length-1)
    result = session.get(url, stream=True)
```

The test code that runs this is here, but you'd have to go a level deeper to see everything going on:
https://github.com/astropy/astroquery/pull/876/files#diff-3f89e9b275405817d06dc7893f6ec23c

nateprewitt:
Thanks @keflavich, so I'm not seeing how you're actually using the `Content-Length` header in the pseudo-code or the linked test. Per the [RFC](https://tools.ietf.org/html/rfc7233#section-4.2), the total length SHOULD be provided in the `Content-Range` header, delimited by a `/` and appears to be on all responses from httpbin. You should also be receiving a 206 if the server is intentionally only sending partial data. 

If you're unintentionally receiving an incomplete response, you'll find that `requests`, which you appear to be using under the hood, should handle incomplete responses by raising an appropriate exception. This is done regardless of `Transfer-Encoding` or `Content-Length` headers. 

#259 is precedent for reverting this since it was decided that the actual server, rather than httpbin, should be emitting the sizing headers in this case. Before we back this out though, I'd like to make sure we try to address your concerns. Could you clarify a bit further why you need a `Content-Length` header here?

keflavich:
@nateprewitt I'll try to clarify more, but note that I edited the example to include `content-length` after writing it; is the example clear(er) now?

nateprewitt:
Ok, that definitely helps. So I think there are a few better ways to accomplish this. By forcing httpbin to return a `Content-Length` header, you're testing httpbin rather than how your code will handle this situation with a server in the wild.

If your code makes it past line 2 (`response = session.get(url, stream=True)`), you've received exactly what the server told you it would send. Requests will raise an exception if the a response with `chunked` encoding is sent incorrectly, or if the `Content-Length` doesn't match what was received.

If your concern is partial responses, I'd suggest doing a check for a 206 response code and then sending a followup request with `headers['Range'] = 'bytes={0}-'.format(existing_file_length)`. This will request everything from your current position to the end and not rely on an optional header.

If you are dealing with a server that arbitrarily sends incomplete data, and doesn't send a 206 as the status code, there's not a lot you can reliably do. The `Content-Length` header, if you get one, probably shouldn't be trusted either. I'd try to contact the administrator of the server at that point because what they're doing is non-sensical. I don't believe this is a state you can hit with your current implementation though.

keflavich:
OK, there seems to be quite a bit about the http returns of partial files I hadn't appreciated.  Thanks for the guidance; I'll try your suggested approach.

Issue #317, opened 2016-12-30
DELETE method is not supposed to support POST form


puiterwijk:
According to the Flask documentation (http://flask.pocoo.org/docs/0.12/api/#flask.Request.form), request.form is only supposed to contain the urlencode parsed form with PUT and POST requests.

This means that the test_delete_endpoint_returns_body test is NOT corresponding to the Flask documentation, and the fact that it DOES work should be considered a bug in Flask.

AdamWill:
For the record, we encountered this when running the httpbin test suite against an older version of Flask (in [EPEL](https://fedoraproject.org/wiki/EPEL) 7). The older version of Flask actually does behave as documented.

puiterwijk:
This works since Werkzeug 0.9, although it's undocumented in Flask: https://github.com/pallets/werkzeug/blob/master/werkzeug/wrappers.py#L328

puiterwijk:
https://github.com/pallets/flask/issues/2132 is the upstream bug about this documentation inconsistency: werkzeug <0.9 only parsed the form for POST, PUT and PATCH requests.

sigmavirus24:
Oh, so y'all are trying to package something modern against the ancient versions that Red Hat supports in its various distributions? Yeah. We're not removing *existing behaviour* to support old versions of Flask and Werkzeug. Sorry.

AdamWill:
Well, sure, but the point is that httpbin's expectations of Flask didn't line up with its documented behaviour. That's an inconsistency no matter how we happened to encounter it. It seems that the behaviour on Flask's part is intentional, and their response was to change the documentation, as covered in https://github.com/pallets/flask/issues/2132 , but that wasn't obvious.

Issue #143, opened 2014-06-08
The body of a DELETE request doesn't show up


subnetmarco:
It's not forbidden by the HTTP spec to have bodies in a `DELETE` request, but executing the following command doesn't return the right body parameters in the response:

```
curl -X DELETE -d "name=Mark" http://httpbin.org/delete
```

Returns:

``` json
{
  "args": {},
  "data": "",
  "headers": {
    "Accept": "*/*",
    "Connection": "close",
    "Content-Length": "9",
    "Content-Type": "application/x-www-form-urlencoded",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.30.0",
    "X-Request-Id": "04955a24-2ccc-4f57-ad3e-7a269c89a2ea"
  },
  "json": null,
  "origin": "208.90.213.84",
  "url": "http://httpbin.org/delete"
}
```

PR #318, opened 2016-12-30
Remove form from /delete endpoint

puiterwijk:
This was never supposed to work according to the Flask documentation.

Fixes: #317
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -183,7 +183,7 @@ def view_delete():
     """Returns DELETE Data."""
 
     return jsonify(get_dict(
-        'url', 'args', 'form', 'data', 'origin', 'headers', 'files', 'json'))
+        'url', 'args', 'data', 'origin', 'headers', 'files', 'json'))
 
 
 @app.route('/gzip')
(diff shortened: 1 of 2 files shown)


AdamWill:
Well, I guess the question is why does httpbin consider this behaviour so important? I guess we should look at why the test was added in the first place, what is the reason that httpbin specifically expects this behaviour from Flask/werkzeug.

@puiterwijk ,  what about this, which you said on IRC but didn't mention here:

"Which makes sense, given that DELETE requests don't even provide a Content-Type application/x-form, and unless that exists, parsing it as a form is at best not according to standards, at worst dangerous"

AdamWill:
OK, so httpbin was explicitly changed to do this in 2593a8ab -

"Add form and files to DELETE Endpoint
    
The DELETE endpoint was missing form and files, so if you hit the endpoint with data in the envelope and a Content-Type header set to application/x-www-form-urlencoded, the data will disappear.  Since RFC 2731 doesn't disallow content in the body of a DELETE request, let's allow that data to pass through
    
Closes #143"

@puiterwijk , what do you think of that?

PR #319, opened 2016-12-30
Use Response.data with older Werkzeug versions

puiterwijk:
Response.get_data did not exist in older versions of Flask/Werkzeug,
and is even now not documented at the Flask level.
This will make httpbin work with either the new version where it does
exist, and the older versions where it doesn't.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


kennethreitz:
✨🍰✨

Issue #344, opened 2017-03-28
/range/ regression


lietusme:
We are using /range/ to simulate slow download, it seems to be no longer transferring chunks, but rather everything in one go at the end of duration:

http://httpbin.org/range/1024?duration=10&chunk_size=2

About 2 weeks ago it was working fine - data was being sent incrementally in chunks.

Any ideas?

nateprewitt:
It looks like b4677d7 changed /range to force a `Content-Length` header. This seems to be preventing the `Transfer-Encoding: chunked` functionality that httpbin previously had which allowed for trickled data.

lietusme:
Does this means its broken or WAD? API allows declaring chunk_size, that does not do anything now

nateprewitt:
I would consider it broken. The endpoint previously streamed data and is now inhibited from providing that functionality. I left a note in #314 regarding reverting this change.

nateprewitt:
I've opened #345 to address this. Pending it's acceptance, we should have this fixed in the next release.

PR #345, opened 2017-03-28
Revert "Return content-length in 'range' requests"

nateprewitt:
Reverts kennethreitz/httpbin#314

@kennethreitz @sigmavirus24 @Lukasa If any of you have a moment, could I get confirmation I'm not completely off base here. #259 made the decision not to include data length related headers in httpbin and allows the server to specify. Requiring a `Content-Length` header for the streamed response seems fundamentally incorrect and has broken portions of the `/range` endpoint.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -623,8 +623,7 @@ def range_request(numbytes):
         response = Response(headers={
             'ETag' : 'range%d' % numbytes,
             'Accept-Ranges' : 'bytes',
-            'Content-Range' : 'bytes */%d' % numbytes,
-            "Content-Length": str(numbytes),
+            'Content-Range' : 'bytes */%d' % numbytes
             })
         response.status_code = 416
         return response
@@ -651,9 +650,7 @@ def range_request(numbytes):
         'Content-Type': 'application/octet-stream',
         'ETag' : 'range%d' % numbytes,
         'Accept-Ranges' : 'bytes',
-        "Content-Length": str(numbytes),
-        'Content-Range' : content_range
-    }
+        'Content-Range' : content_range }
 
     response = Response(generate_bytes(), headers=response_headers)
 


lietusme:
This escalated from #344 - data was not being streamed from /range, but returned all in one go at the end. Looking from API standpoint response headers could contain Content-Length and still return data incrementally - there should be no standard violation here. Is server implementation preventing that somehow?

kennethreitz:
I defer to @nateprewitt and @Lukasa here.

nateprewitt:
@Lukasa I definitely agree that `Content-Length` isn't forbidden. In fact, it's even used in the RFC spec examples for Range. What I'm trying to rectify here is the fact `httpbin` is now unconditionally forcing a `Content-Length` header. On top of that, the value isn't correct for any request using a range header other than `Range: 0-` because the content length is always the full length passed to `/range/<num bytes>`.

It seems the requirements that prompted this PR in the first place may have not been fully fleshed out, making this change unnecessary. Since it doesn't seem to be providing significant value and is actively affecting the endpoint integrity, I'm suggesting we revert it.

@lietusme My local tests yesterday *seemed* to suggest the `Content-Length` was affecting the streaming. I'm getting mixed results locally now when watching things with tcpdump. It's very possible there's an issue in the new deployment on Heroku. Either with gunicorn or heroku?

Lukasa:
I have no objection to removing it, but the other option would surely be to fix it so it returns the correct length, yes?

nateprewitt:
@Lukasa, if we want to adopt the forced use, yeah. I figured this was the least intrusive to how things worked prior to #314. If there's a good reason to adopt `Content-Length` on every range request then I agree we should just fix it.

Lukasa:
I'm kind of inclined to say that we should. Not putting it there is a bit gross, really.

lietusme:
I aggree that Content-Lenght makes sense in /range responses:
- Length is known by service - thats by standard to return if known
- Range API purpose is ranged download, not streaming service

That said, /range should still return chunked response incrementally. 

@nateprewitt - no idea if its server/deployment issue. I currently use https://httpbin.org, planning on deploying locally in future.

Lukasa:
> That said, /range should still return chunked response incrementally.

Hang on, we need to clarify something.

If range is returning chunked, it *must not* set Content-Length. RFC 7230 forbids it. By exclusion, then, if it's setting Content-Length, it *must not* return chunked. You can have either: not both.

lietusme:
Sorry, I did not mean Chunked Transfer Encoding, but the fact that /range could still trickle transfer incrementally for given duration in given chunk sizes - simulating slow connection. This does not mean that content should be undefined length, but just that its being transferred continuously. This is my primary use of /range API - simulating slow connection. Maybe my workflow needs different API, but usecase seemed appropriate for what /range does.

Lukasa:
Yup, so that can definitely work.

The general rule for WSGI is that when an application returns a Content-Length header the WSGI server will *not* chunk the data: it'll just enforce the content length. When the WSGI app does not return Content-Length, the server may choose to either set Connection: close or Transfer-Encoding: chunked, depending on its own preferences.

In this case, we know the Content-Length, so we should just set it and be done.

nateprewitt:
Ok, I'm on board. I've added #346 to supersede this.

PR #346, opened 2017-03-30
fix Content-Length header values for /range endpoint

nateprewitt:
This supersedes #345 by fixing the value for the Content-Length for the `/range` endpoint.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -618,13 +618,14 @@ def range_request(numbytes):
 
     request_headers = get_headers()
     first_byte_pos, last_byte_pos = get_request_range(request_headers, numbytes)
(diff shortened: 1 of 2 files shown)


nateprewitt:
Great, I'll go ahead and merge this then. Thanks @Lukasa!

lietusme:
Thanks, when do you usually deploy changes to httpbin.org ?

kennethreitz:
promoted to production!

lietusme:
Hi, this is still working as previously on httpbin.org (response sent at the end of duration, not streamed incrementally). Not sure if question should go here, but when this will be fixed in httpbin.org site as well?

PR #350, opened 2017-04-22
Adding a /etag URL for testing If-Match and If-None-Match conditional logic

mmattozzi:
I thought it would be interesting if a client could supply an entity tag in the URL and then httpbin would execute the appropriate conditional logic based on the request's `If-Match` and `If-None-Match` headers. 

Some examples in action: 
```
$ curl -I -H 'If-None-Match: "abc"' 'http://localhost:8080/etag/abc'
HTTP/1.1 304 NOT MODIFIED

$ curl -I -H 'If-None-Match: "abcd"' 'http://localhost:8080/etag/abc'
HTTP/1.1 200 OK

$ curl -I -H 'If-Match: *' 'http://localhost:8080/etag/abc'
HTTP/1.1 200 OK

$ curl -I -H 'If-Match: "abc"' 'http://localhost:8080/etag/abc'
HTTP/1.1 200 OK

$ curl -I -H 'If-Match: "efg"' 'http://localhost:8080/etag/abc'
HTTP/1.1 412 PRECONDITION FAILED

$ curl -I -H 'If-None-Match: "abc", "d", "efg"' 'http://localhost:8080/etag/efg'
HTTP/1.1 304 NOT MODIFIED
```

Since usage of both headers at once is indeterminate in the spec, `If-None-Match` is preferred. Modification date is not considered, I just assume the behavior would be as if the last modified date is the present time. 

Let me know what you think!

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -23,7 +23,7 @@ from werkzeug.wrappers import BaseResponse
 from raven.contrib.flask import Sentry
 
 from . import filters
-from .helpers import get_headers, status_code, get_dict, get_request_range, check_basic_auth, check_digest_auth, secure_cookie, H, ROBOT_TXT, ANGRY_ASCII
+from .helpers import get_headers, status_code, get_dict, get_request_range, check_basic_auth, check_digest_auth, secure_cookie, H, ROBOT_TXT, ANGRY_ASCII, parse_multi_value_header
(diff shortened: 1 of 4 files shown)


kennethreitz:
✨🍰✨

kennethreitz:
deployed to production.

PR #354, opened 2017-05-11
Mention that "stream-bytes" uses chunked encoding

janko-m:
I needed an endpoint which returns content in chunked encoding (`Transfer-Encoding: chunked`), and through `curl`-ing around I found out that `stream-bytes` does that, so I thought it would be useful to document it.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #358, opened 2017-05-26
add POST to methods kwarg

### LINKED ISSUE #357: Allow for POSTs to /response-headers
I have a test case I would like to leverage httpbin for that involves making a POST to a page and expecting a particular header in order to proceed with the request chain. Unfortunately, the current endpoint only expects GETs.

Would it be possible to allow POSTs and merge the submitted headers (ala /headers) with the POST data?

I am happy to make this change but wanted to make sure it was a supported idea. Thanks in advance.

mattwwarren:
and a test, for good measure

Fixes #357

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -337,7 +337,7 @@ def view_status_code(codes):
     return status_code(code)
 
 
-@app.route('/response-headers')
+@app.route('/response-headers', methods=['GET', 'POST'])
 def response_headers():
     """Returns a set of response headers from the query string """
     headers = MultiDict(request.args.items(multi=True))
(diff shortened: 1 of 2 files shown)


httpbin/core.py
```diff
@@ -337,7 +337,7 @@ def view_status_code(codes):
     return status_code(code)
 
 
-@app.route('/response-headers')
+@app.route('/response-headers', methods=['GET', 'POST'])
```
mattwwarren: While this is the simplest change, I realize it is functionally odd because you're submitting a POST to a path with a query string. I may add to this to use both the data and the args.

kennethreitz:
this seems sane to me!

PR #360, opened 2017-06-01
Add /anything route

Flimm:
This makes it easier to test which HTTP method is being used, and also to test custom URLs.

Unfortunately, it isn't good for testing URLs with special Unicode characters in them, because Flask seems to automatically decode and unescape the URLs.


kennethreitz:
i like it!

kennethreitz:
i'll try to deploy this into production tonight :D

kennethreitz:
http://staging.httpbin.org

PR #364, opened 2017-06-25
Digest authentication now can report nonce stale

MarekR22:
New functionality is needed to test scenario where server decides that digest nonce value provider by server is not valid any more.
In such case server always responds with HTTP code 401 no mater if user password is valid or not and WWW_Authenticate header contains "stale=true".

In provided change digest url is made longer by extra parameter "stale_after" which indicates number of requests with valid credentials before nonce expiration is reported.


MarekR22:
I have improved version with propr test written so will create new pull request and close this one.

PR #366, opened 2017-07-08
Digest authentication and expiried nonce - stale=true scenerio

MarekR22:
Hi,

In this change digest authentication path is expanded by extra value `stale_after`.
This integer value will cause that after `stale_after` successful responses next response will report `401` with flag `stale=true` indicating that nonce value is no longer valid.

Also each `401` response stores last used `nonce` value in cookie, so if another request is received with same nonce value response with status code `401` and with flag `stale=true` is send back even if password is correct (nonce is considered as expired).

I've also refactor test a bit and added a new test to cover new functionality and added extra one for invalid credentials.

Hope you will consider this as useful change.

BR,

Marek

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -482,6 +482,75 @@ def digest_auth(qop=None, user='user', passwd='passwd', algorithm='MD5'):
         return response
     return jsonify(authenticated=True, user=user)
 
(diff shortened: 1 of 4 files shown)


kennethreitz:
Merged! ✨🍰✨

kennethreitz:
Deployed to production!

PR #367, opened 2017-07-14
each 'id' should only appear once

fanofxiaofeng:
h2 tags in line 5 and line 11 both have id 'ENDPOINTS', I suppose the h2 tag  in line 5 has a wrong id.


kennethreitz:
thanks!

Issue #361, opened 2017-06-07
"/anything" route only allows access by GET method


warren-bank:
```bash
> curl -X POST -d 'a=1&b=2' https://httpbin.org/anything

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

```bash
> curl -X PUT -d 'a=1&b=2' https://httpbin.org/anything

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

notes:
* [code](https://github.com/kennethreitz/httpbin/blob/3463b5ec56795a32ac44bb3e6af8ff720198ce10/httpbin/core.py#L167)
* should also output: 'data'

warren-bank:
maybe something like..

```python
@app.route('/anything', methods=['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'TRACE'])
@app.route('/anything/<path:anything>', methods=['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'TRACE'])
def view_anything(anything=None):
    """Returns request data."""

    return jsonify(get_dict('url', 'args', 'headers', 'origin', 'method', 'form', 'data', 'files', 'json'))
```

OJFord:
Because the `/:verb` routes don't allow path data, (e.g. `/get/123`) the combined effect with this issue is that there's no way (that I can see) to use httpbin to simulate a RESTy 'collection'.

If the `/anything` route allowed all methods, then one could `POST /anything`; `GET /anything/123`; DELETE `/anything/123`; etc. (Note I'm not suggesting that httpbin actually create resources, that's clearly out of its scope.)

OJFord:
> Because the /:verb routes don't allow path data, (e.g. /get/123) the combined effect with this issue is that there's no way (that I can see) to use httpbin to simulate a RESTy 'collection'.

Wrong: `/status/:code` allows all methods, so valid HTTP status codes can be used as resource IDs.

warren-bank:
[...]

PR #369, opened 2017-07-19
Allow all methods on /anything

macheins:
`/anything` already returns the used method, but only allows for `GET` requests.
With this extension `/anything` acts as an "echoserver" and additionally returns any sent data.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -164,12 +164,12 @@ def view_get():
     return jsonify(get_dict('url', 'args', 'headers', 'origin'))
 
 
-@app.route('/anything')
-@app.route('/anything/<path:anything>')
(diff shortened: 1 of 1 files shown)


macheins:
Also fixes #361

MarekR22:
based on description it would be nice to rename to "/anything" to "/echo".

kennethreitz:
thanks!

dair-targ:
While filing #522 I was surprised that according to the standard much more HTTP methods exist beyond ones mentioned in this merge request. Is there a strong preference, or I'll just submit a patch to allow an arbitrary methods at `/anything`?

jimmywarting:
Came here looking/suggesting for any method on "/anything" also

would like to be able to do any custom http method

PR #372, opened 2017-07-29
Add env endpoint

maruina:
The `/env` endpoint takes as parameter an environment variable name and returns its value or 404 if the variable doesn't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ dist/
 .tox
 *.egg-info
 *.swp
+
+.idea
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -778,6 +778,15 @@ def xml():
     return response
 
 
+@app.route("/env/<env_var>")
+def env(env_var):
+    value = os.getenv(env_var)
+    if value is None:
+        return status_code(404)
+    else:
+        return jsonify({env_var: value})
+
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument("--port", type=int, default=5000)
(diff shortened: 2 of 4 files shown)


Lukasa:
This strikes me as an *extremely* bad idea to add to the httpbin application: it runs the risk of enabling remote users to exfiltrate data about the deployed environment containing httpbin. Given that httpbin instances are frequently run on publicly reachable hosts, it strikes me as a pretty severe information leak to add this endpoint.

maruina:
The use case for this endpoint, is to be able to test different "versions"
of the same application. This is particular useful when you want to test
routing rules, or load balancing between multiple deployment.

What other options would you suggest? Would be ok to have an endpoint
called version to read only that environment variables? Would you prefer
whitelisting a couple of vars like version and environment?

Lukasa:
Different versions of what application? httpbin?

maruina:
Yes, the idea is that if you want to test routing rules (for example, a  blue-green deployment in a cluster), instead of actually deploying two different versions of the app, you can set an environment variable with two different value.

My recent use case for that was while I was testing the [Istio routing](https://istio.io/docs/tasks/request-routing.html) functionality. In the [example](https://istio.io/docs/samples/bookinfo.html), they deploy three version of the same application. Instead of doing that, I found useful deploying the same application and using the same endpoint for  getting different values.

sigmavirus24:
So I understand the desire, @maruina, but it's not practical or secure for httpbin. Instead a `metadata` endpoint that returns static information would be better but I do worry that this project is starting to experience endpoint sprawl.

Lukasa:
I also think, ultimately, that it's out of scope. It has nothing to do without HTTP as a protocol and everything to do with infrastructure. I recommend either deploying a forked copy, or writing a minimal flask app that can identify itself in this way

PR #387, opened 2017-09-24
Split content into paragraphs

vanillajonathan:


kennethreitz:
Let's not change this. people might be relying on it.

PR #392, opened 2017-10-15
Add back fake cookie and make requiring it conditional

### LINKED ISSUE #380: fake_value cookie removed, which breaks a test in requests
This commit: 791e31cba4a4545352fa7d6e2d7019910166160a  removed the cookie that was always set on the digest endpoint, which breaks a test in requests (see also: [this travis failure]( https://travis-ci.org/requests/requests/jobs/268971293))

There was some old (never-resolved) discussion of this odd feature here: kennethreitz/httpbin#124

Paging @sigmavirus24 - should we add this behavior back?

See also: requests/requests#4263

sigmavirus24:
Our digest auth endpoint will once again return a fake cookie but will
only require it if the query-string parameter `require-cookie` is
present.

Further, this fixes a bug in the digest auth implementation in httpbin
where we were only considering the path portions of the request-uri in
the specification. RFC 7230 is very clear that the request-uri includes
the query-string if it is present which means it is necessary in our
digest auth handling.

Closes #380

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -467,44 +467,58 @@ def digest_auth_nostale(qop=None, user='user', passwd='passwd', algorithm='MD5')
 @app.route('/digest-auth/<qop>/<user>/<passwd>/<algorithm>/<stale_after>')
 def digest_auth(qop=None, user='user', passwd='passwd', algorithm='MD5', stale_after='never'):
     """Prompts the user for authorization using HTTP Digest auth"""
+    require_cookie_handling = (request.args.get('require-cookie', '').lower() in
+                               ('1', 't', 'true'))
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This should also fix Requests' test suite.

sigmavirus24:
I have no clue why @kennethreitz's Heroku account is having so much trouble deploying a review app for this but :shrug:

@kevin1024 what do I need to do to get pytest-httpbin to craft a release using this?

kevin1024:
I don’t believe you need to do anything to pytest-httpin.  You should just need to do a new release of httpbin itself.

sigmavirus24:
✨ Released ✨

PR #393, opened 2017-10-20
Add support for basic auth without a password

### LINKED ISSUE #356: Basic auth with no password
Basic auth is sometimes used where the username is an API key, and the password is blank.

Is there a way to make httpbin act in this way? I can see `https://httpbin.org/basic-auth/user/passwd`, would it be possible to also have `https://httpbin.org/basic-auth/user` for cases when there is no password.

I would then expect both of the following to work:

```
curl -u user:passwd https://httpbin.org/basic-auth/user/passwd
curl -u user: https://httpbin.org/basic-auth/user
```

Thanks for this project, it's great. I dont use python day-to-day but i'm happy to submit a pull request for this if you think it's a good idea.

domharrington:
Finally got round to working on this. Fixes #356.

It's my first time writing python, so let me know if the style isn't up to scratch. Also added some tests for basic auth because i couldnt see them anywhere else, but if i missed them and i've duplicated it, let me know.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -435,6 +435,16 @@ def delete_cookies():
     return r
 
 
+@app.route('/basic-auth/<user>')
+def basic_auth_without_password(user='user'):
+    """Prompts the user for authorization using HTTP Basic Auth without a password."""
+
+    if not check_basic_auth(user, ''):
+        return status_code(401)
+
+    return jsonify(authenticated=True, user=user)
+
+
 @app.route('/basic-auth/<user>/<passwd>')
 def basic_auth(user='user', passwd='passwd'):
     """Prompts the user for authorization using HTTP Basic Auth."""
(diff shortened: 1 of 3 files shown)


kennethreitz:
this looks solid, but a bit silly. i'm torn, personally.

sigmavirus24:
@kennethreitz I agree that this seems silly. That said, httpbin doesn't seem to have a strict philosophy. Is it there to test the most common variations of things that people will see, or is there to allow folks to test everything? If we can decide on that, I think we can have clearer answers for different additions to the project.

kennethreitz:
i think i'll opt against another endpoint, but thank you for your kind contribution! it's a sane one and a good one. just, unneccessary. reducing code coverage is always a good thing.

kennethreitz:
✨🍰✨

PR #401, opened 2017-11-19
add SHA-512 authentication

linuxlizard:
Add SHA-512 authentication alongside SHA-256 (RFC-7616)


nateprewitt:
Not sure if we need a release cut to get this added to pytest-httpbin. @sigmavirus24?

linuxlizard:
A pytest-httpbin release would be handy for adding my test code into Requests for the same SHA-256, SHA-512 feature.

PR #404, opened 2017-11-26
Updated tox and Travis to use py36.

javabrett:
Update Python interpreter used in Tox tests from py34 to py36.  3.6 seems to be what is used in the current Docker container.  Not sure what is deployed to sites.

Made equivalent changes to Travis config.  FWIW also locked dist to trusty, and added a matrix to split the 2.7 and 3.6 builds to use exactly those runtimes (possible the tox env was achieving this anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
+dist: trusty
 language: python
 
 sudo: false
-env:
-    - TOXENV=py27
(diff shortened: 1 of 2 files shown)


kennethreitz:
thanks for merging this, @sigmavirus24

Issue #340, opened 2017-03-20
Regression with chunked transfer encoding?


pb-:
I'm using https://httpbin.org and it looks like chunked transfer encoding does not work any more as expected:

```
$ curl -X POST https://httpbin.org/post \
 -d '{"message":"BLA"}' \
 -H 'Content-Type: application/json' \
 -H 'Transfer-Encoding: chunked'
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
    "Connect-Time": "1", 
    "Connection": "close", 
    "Content-Type": "application/json", 
    "Host": "httpbin.org", 
    "Total-Route-Time": "0", 
    "Transfer-Encoding": "chunked", 
    "User-Agent": "curl/7.50.1", 
    "Via": "1.1 vegur", 
    "X-Request-Id": "2a06bd5b-06dc-4ced-ac89-fe83cbb6771c"
  }, 
  "json": null, 
  "origin": "<removed>", 
  "url": "https://httpbin.org/post"
}
```

Observe that `data` and `json` do not contain the posted data. The same request works as expected without chunked transfer encoding:

```
$ curl -X POST https://httpbin.org/post \
 -d '{"message":"BLA"}' \
 -H 'Content-Type: application/json'
{
  "args": {}, 
  "data": "{\"message\":\"BLA\"}", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
    "Connect-Time": "0", 
    "Connection": "close", 
    "Content-Length": "17", 
    "Content-Type": "application/json", 
    "Host": "httpbin.org", 
    "Total-Route-Time": "0", 
    "User-Agent": "curl/7.50.1", 
    "Via": "1.1 vegur", 
    "X-Request-Id": "dfbaa440-9871-4a3c-b482-35a68459722a"
  }, 
  "json": {
    "message": "BLA"
  }, 
  "origin": "<removed>", 
  "url": "https://httpbin.org/post"
}
```

Am I missing something?

sigmavirus24:
You're not missing anything. This does look like a regression.

For @kennethreitz:

```py
>>> r = requests.post('https://httpbin.org/post', data=iter(('foo', 'bar', 'bogus')))
>>> r.request.headers['Transfer-Encoding']
>>> r.json()['data']
>>> pprint.pprint(r.json())
{u'args': {},
 u'data': u'',
 u'files': {},
[...]

PR #405, opened 2017-11-27
Added support for chunked encoded requests when running gunicorn. #340

javabrett:
In this commit:

- when we see a Transfer-Encoding: chunked request, and the server is gunicorn,
  we set environ wsgi.input_terminated, which is required by Werkzeug in the
  absence of Content-Lenght, or it will empty the data stream.
- for chunked requests to non-gunicorn, return 501 Not Implemented.

Looking for guidance on the best way to set-up an acceptance/system test for this, since it cannot be verified through unit tests using FlaskClient.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -101,7 +101,7 @@ def before_request():
         server = request.environ.get('SERVER_SOFTWARE', '')
         if server.lower().startswith('gunicorn/'):
             if 'wsgi.input_terminated' in request.environ:
-                app.logger.warning("environ wsgi.input_terminated already set, keeping: %s"
+                app.logger.debug("environ wsgi.input_terminated already set, keeping: %s"
                                    % request.environ['wsgi.input_terminated'])
             else:
                 request.environ['wsgi.input_terminated'] = 1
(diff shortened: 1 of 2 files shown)


httpbin/core.py
```diff
@@ -84,6 +84,30 @@ def jsonify(*args, **kwargs):
...
+- flask will hang and does not seem to properly terminate the request, so
+  we explicityly deny chunked requests.
+"""
+@app.before_request
+def before_request():
+    if request.environ.get('HTTP_TRANSFER_ENCODING', '').lower() == 'chunked':
+        server = request.environ.get('SERVER_SOFTWARE', '')
+        if server.lower().startswith('gunicorn/'):
+            if 'wsgi.input_terminated' in request.environ:
+                app.logger.warning("environ wsgi.input_terminated already set, keeping: %s"
```
sigmavirus24: Is this really a warning level message? I'd consider this `INFO` or `DEBUG` at best.

javabrett: Yes `debug` is better and I have changed to that and rebased to master.

The original thinking of a louder log message was to quickly catch a change in `gunicorn` if it started adding `input_terminated` to requests, deprecating this middleware.  But debug is fine for this.

kennethreitz:
✨🍰✨

PR #412, opened 2017-12-17
Rewrite of core and tests

Singletoned:
I’ve been working on [Requests-WSGI-Adapter][1] so that I can use Requests for testing WSGI apps, but I needed it to work on quite a few different versions of Python (it’s to help me test legacy apps as the first step to getting them off old versions of Python).  I decided it would be a good idea to use HttpBin to test Requests-WSGI-Adapter to make sure that it was handling everything that it should.  Unfortunately, because of Requests’ usage of Flask, it didn’t install on many older versions of Python.  I [forked HttpBin][2] and rewrote it to use just Werkzeug, and removed all the other Flask dependencies.  I also altered the tests to use Requests-WSGI-Adapter and Requests to test a HttpBin WSGI app as if it was just a remote site (which means the tests can be run against a live version of the site as well).  I also significantly expanded the test suite (as I needed to test all the aspects of Requests-WSGI-Adapater).

I think it is of interest as an example of how easy it is to build a small web app with libraries instead of frameworks (which is a bit of thing of mine). (The other thing of mine is that testing libraries have to support Python 2.6, otherwise large legacy apps in corporations will never get off it, and the management will feel they have been “burned” by Python).

I also think it could potentially be replacement for HttpBin.  It has better test coverage and is simpler, but there are risks in that I might have accidentally changed undocumented features, etc.  I’m happy to maintain it as a fork anyway, as it serves a purpose for me, but I thought I would bring your attention to it in case it was of interest.

I think at a minimum this would need sanity checking, particularly around untested features.  For example there was a tracking code that I may have messed up.  Unfortunately the diffs of the core and tests are huge.  I have moved over any recent changes since I originally forked the code (such as the SHA-512 change).

I'm happy to discuss this, listen to suggestions, and adapt it how ever people see fit.

Thanks

Ed

[1]: https://github.com/Singletoned/requests-wsgi-adapter
[2]: https://github.com/Singletoned/httpbin


sigmavirus24:
I'll let @kennethreitz weigh in, but I don't think we are going to accept random rewrites that add maintenance burden by forcing the actual maintainers to:

- Learn something new (in this case, werkzeug) for debugging purposes
- Support a long-dead (seriously, it's been over 4 years now) version of Python. If corporations feel burned by "Python" for having a well published, long-established release schedule that properly publicizes when each version will end support, I don't think we will satisfy them. Especially since HTTPbin is not critical path software for anything of substance.
- Use *deprecated and unsupported* software to maintain the project (`nosetests` was deprecated long ago).

If you want to fork httpbin to accomplish the above goals, you're welcome to do so. Again, this is my opinion. I'll let @kennethreitz have the last word.

kennethreitz:
I'm not against some of these ideas — for example switching from Flask to Werkzeug.

But this pull request is doing far too much in one swoop. 

If you want to contribute this way into this project, why don't you make a bulleted list of the proposed changes for us, and I'll tell you which ones to PR us with :)

Issue #415, opened 2017-12-18
HTTP 500 errors for digest auth


jeroen:
As of earlier today my http client unit tests started failing because the `digest-auth` endpoint returns HTTP 500 instead of 200 for correct credentials. Example:

```
curl https://eu.httpbin.org/digest-auth/auth/jerry/secret -v -u jerry:secret
```

Has there been a new deployment? It was working OK for the past months, maybe even years.

sigmavirus24:
This could be related to https://github.com/kennethreitz/httpbin/commit/9e9cb84724cfb72e7780a7c37c59468e55bb8f0d but it's unclear to me if @kennethreitz has deployed recently

sigmavirus24:
Looking a little closer I see:

```
*  SSL certificate verify ok.
* Server auth using Basic with user 'jerry'
> GET /digest-auth/auth/jerry/secret HTTP/1.1
> Host: eu.httpbin.org
> Authorization: Basic amVycnk6c2VjcmV0
> User-Agent: curl/7.54.0
> Accept: */*
```

From the `curl` output. It seems like httpbin was incorrectly accepting basic auth credentials for digest auth. We should investigate more to understand what the behaviour was previously.

jeroen:
Yes that is correct. The problem is it now returns a `HTTP 500` (which indicates a server bug).

What should be happening, and was happening before, is that the server ignores the basic auth and returns `HTTP 421 UNAUTHORIZED` along with the proper `www-authenticate` header according to the http digest protocol. This invites the client to make a subsequent request including the proper `Authorization` header.

This does work:

```
curl https://eu.httpbin.org/digest-auth/auth/jerry/secret -v -u jerry:secret --digest
``

But really this should also work (and it used to):

```
curl https://eu.httpbin.org/digest-auth/auth/jerry/secret -v -u jerry:secret
```

The client doesn't have to know in advance that the server wants to use digest auth. If it first tries to authenticate basic auth the server should ignore the basic auth credentials and continue with the digest auth dance, not just error.

**Comment by sigmavirus24** [comment]
Right, I'm not arguing that it should 500. I agree that's wrong. I'm trying to discern what the default behaviour *was*

**Comment by jeroen** [comment]
OK cool. AFAIK the previous (and correct) behaviour for the digest auth API is simply to ignore the basic auth header and respond identically as when no basic auth header was sent.

**Comment by jeroen** [comment]
I reproduced this with the official httpbin docker image. Below a stack trace:

```
[...]

PR #424, opened 2018-01-28
Send digest-auth challenge for wrong Authorization header

nateprewitt:
This should resolve the issue in #415. We had a regression around when we send digest-auth challenges with #366 and later #392. This will now verify what kind of 'Authorization' header we received and send a challenge if the value isn't `digest xxxxxxxxxx`.


nateprewitt:
Alright, this is rebased onto master and tests are passing now.

jeroen:
Thanks! Is this deployed immediately or not yet?

PR #425, opened 2018-01-28
Remove test for empty Content-Type

nateprewitt:
Werkzeug 0.14.0 introduced new behaviour stripping falsey values from being included in EnvironHeaders ([625362a](https://github.com/pallets/werkzeug/commit/625362a429263f8d713a9ed06c18f247042d7d0f)). We were previously testing `Content-Type` header was empty which is now failing. From local tests with older versions of werkzeug, it looks like we weren't emitting the empty header in the /get or /anything responses.

I think we're *probably* OK removing these checks since they didn't seem to have had any production implications.


sigmavirus24:
Would it make sense to add `14.0.0` as the lowest possible version?

nateprewitt:
I'd say probably not since `werkzeug` isn't a direct dependency of `httpbin`. If we start requiring a specific version but leave Flask (which is what's currently installing it) unpinned, that may cause issues. Since we "don't see user facing impact"™ based on version, I think we should probably leave it up to the users' install of Flask to decide its dependency requirements.

sigmavirus24:
> If we start requiring a specific version but leave Flask (which is what's currently installing it) unpinned

Having `werkzeug >= 0.14.0` and `Flask >= whatever` wouldn't be objectionable to me.

> I think we should probably leave it up to the users' install of Flask to decide its dependency requirements.

If folks are both using httpbin and Flask in a project, I'd be surprised and willing to rollback the lower version restrictions.

I suspect if there was a test encapsulating something here then we were testing it for some reason.

nateprewitt:
From the details in #173 it looks like this was just testing an implementation detail, but like you said we can rollback if needed.

Pin for werkzeug added, couldn't find a modern version of Flask that it wasn't working with.

PR #426, opened 2018-02-05
document decode_base64 endpoint /base64/:value on the index.html

tmack8001:
I have come to find out that the /base64/:value endpoint isn't documented on the index.html page, I'm suggesting adding this into that page.


sigmavirus24:
Well done and thank you! ✨ 🍰 ✨

Issue #442, opened 2018-04-30
Flask-Common not compatible with Flask-1.0


nateprewitt:
Requests test suit broke a couple hours ago with the release of Flask-1.0. The flask-common module in httpbin relies on the `flask.exthook` module which was removed, so we now get import errors. We can either update `flask-common` which seems to have been moved into read-only mode on Github, we pin `Flask<1.0`, or we can remove it as a dependency.

@kennethreitz do you have any opinion on this? If it's not going to be maintained, I propose we do a one time update to be tolerant of Flask-1.0 changes, and then move to deprecate.

5j9:
BTW, the issue has occurred after `Flask-1.0.1` release, `Flask-1.0` seems to be OK. See https://github.com/pallets/flask/commit/715a9a3e7bd65421c36428ba9d4a1ee06b98bbc1.

javabrett:
I suggest pinning to `Flask-0.12.2`, which is known-compatible and what was used prior to the 1.0/1.0.1 Flask releases.  This restores sanity for now and rights the build on master while decisions per above are made.

PR #445, opened 2018-05-02
Locked Flask version to <1.0.0

javabrett:
Temp workaround for #442 whilst compat with Flask 1.0.1 is worked-on.


sigmavirus24:
I think flask common is the issue and it seems to be no longer actively developed. Let’s remove it as a dependency

PR #447, opened 2018-05-05
Removed Flask-Common. Fixed #442.

### LINKED ISSUE #442: Flask-Common not compatible with Flask-1.0

javabrett:


Pipfile.lock
```diff
@@ -1,7 +1,20 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e2d593192acadabf3463785be0dc3a1e163d41713ca8d4cd53d4e99c03d927a4"
+            "sha256": "55b1275da9acdf1b325c78f2be89df79f5435330ef7333b8caf9535350c24a72"
+        },
+        "host-environment-markers": {
```
sigmavirus24: I don't think we want these. Can you regenerate the Pipfile.lock without them?

javabrett:
`Pipfile.lock` was updated via `pipenv lock`. According to https://github.com/pypa/pipenv/issues/753 (re)generation of `host-environment-markers` is intended and ongoing.

Is there some switch for `pipenv` to avoid regeneration, or should `pipenv` be avoided, or the entries removed by hand, or https://github.com/pypa/pipenv/issues/753 re-raised?

nateprewitt:
@sigmavirus24, unless it’s been added in the last month or so, I don’t believe there’s any way to disable those.

@javabrett, you’ll want to leave it for now I think. The Pipfile.lock shouldn’t be modified by hand. We’ve typically had a maintainer generate the lockfile to keep some level of consistency. I don’t see a huge reason that’s necessary though, outside of someone else’s system info being stored.

javabrett:
I could just revert the `Pipfile.lock` and let a maintainer lock, or feel free to edit my commit for that, then a maintainer can lock.

I am fairly new to `pipenv` and have been noticing a fair bit of diff churn noise when locking e.g. order of elements.

javabrett:
I since discovered https://github.com/pypa/pipenv/commit/ecf78424c1740bae3511c731c032a3768154072c , and that my `pipenv` was as-shipped with my CPython and hopelessly out-of-date, so upgraded:

```
$ which pipenv
/Library/Frameworks/Python.framework/Versions/3.6/bin/pipenv

$ pipenv --version
pipenv, version 8.3.2

$ pip install -U pipenv

$ pipenv --version
pipenv, version 11.10.1
```

... and `pipenv lock` no-longer adds that host env.

@sigmavirus24 @nateprewitt please re-review.

Issue #403, opened 2017-11-25
0.6.2: self test failure: `static/` missing


0-wiz-0:
When running the self tests for the pypi distfile of 0.6.2 on NetBSD 8.99.7/amd64 with python-3.6.3, I see the following test failure:
````
httpbin (unittest.loader._FailedTest) ... ERROR

======================================================================
ERROR: httpbin (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: httpbin
Traceback (most recent call last):
  File "/usr/pkg/lib/python3.6/unittest/loader.py", line 462, in _find_test_path
    package = self._get_module_from_name(name)
  File "/usr/pkg/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/scratch/www/py-httpbin/work/httpbin-0.6.2/httpbin/__init__.py", line 3, in <module>
    from .core import *
  File "/scratch/www/py-httpbin/work/httpbin-0.6.2/httpbin/core.py", line 61, in <module>
    common = Common(app)
  File "/usr/pkg/lib/python3.6/site-packages/flask_common.py", line 95, in __init__
    self.init_app(app)
  File "/usr/pkg/lib/python3.6/site-packages/flask_common.py", line 112, in init_app
    app.wsgi_app = WhiteNoise(app.wsgi_app, root=url_for('static', filename='')[1:])
  File "/usr/pkg/lib/python3.6/site-packages/whitenoise/base.py", line 61, in __init__
    self.add_files(root, prefix)
  File "/usr/pkg/lib/python3.6/site-packages/whitenoise/base.py", line 96, in add_files
    self.update_files_dictionary(root, prefix)
  File "/usr/pkg/lib/python3.6/site-packages/whitenoise/base.py", line 101, in update_files_dictionary
    stat_cache = dict(scantree(root))
  File "/usr/pkg/lib/python3.6/site-packages/whitenoise/scantree.py", line 21, in scantree
    for entry in scandir(root):
FileNotFoundError: [Errno 2] No such file or directory: 'static/'

````

sigmavirus24:
I don't see this failure. How're you running the tests?

sigmavirus24:
[...]

PR #448, opened 2018-05-06
Added Travis job to run 'python3.6 setup.py test' instead of tox

javabrett:
This adds a third build to the Travis matrix, to run `python3.6 setup.py test` instead of tox.  
This will catch installation issues of the type seen in #403.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ matrix:
           env: TOXENV=py27
         - python: 3.6
           env: TOXENV=py36
+        - python: 3.6
+          script:
+              - python3.6 setup.py test
 
 install:
     - travis_retry pip install tox


sigmavirus24:
I'd rather not. setuptools commands are janky and often end up necessitating ridiculous amounts of debugging when something goes awry. No one presently involved in this project has the time or interest to do that. Also `tox` should be installing `httpbin` via the `setup.py` each time. That makes me skeptical that this would catch the issue.

javabrett:
> Also `tox` should be installing `httpbin` via the `setup.py` each time.

Is it currently only doing that for `release` builds?:

https://github.com/requests/httpbin/blob/master/tox.ini#L14
https://github.com/requests/httpbin/blob/master/tox.ini#L5

```
[testenv]
commands=python test_httpbin.py

[testenv:release]
...
commands =
    python setup.py sdist bdist_wheel
...
```
I'm not familiar with `tox` much, but maybe that is why it didn't catch this between releases ... was running `python test_httpbin.py`, should it be running `python setup.py`

PR #449, opened 2018-05-06
Added version header as X-Powered-By: httpbin/<version>

### LINKED ISSUE #431: Idea: Expost httpbin version from HTTP headers
It would be useful if you could tell the httpbin version from a request's HTTP headers.
This way I would be able to determine why my tests work locally, but not in CI.

javabrett:
This PR adds a HTTP response header `X-Powered-By` containing `httpbin` and its version:

```
Server: gunicorn/19.8.1
X-Powered-By: httpbin/0.6.2
```

This might have been done using multiple products in the `Server` header, but I suspect that a lot of upstream reverse-proxies etc., perhaps `nginx`, `gunicorn` will just overwrite what is written there.

Fixed #431.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -14,6 +14,7 @@ import random
 import time
 import uuid
 import argparse
+import pkg_resources
 
 from flask import Flask, Response, request, render_template, redirect, jsonify as flask_jsonify, make_response, url_for, abort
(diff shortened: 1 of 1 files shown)


kennethreitz:
this would require httpbin to be installed, which it is not in production.

PR #457, opened 2018-05-19
Fix #454: Support `show_env` param in `/headers`

### LINKED ISSUE #454: /headers does not show Via header
```
curl -s -H x-via:bar -H via:foo 'https://httpbin.org/headers' | jq .
{
  "headers": {
    "Accept": "*/*",
    "Connection": "close",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.54.0",
    "X-Via": "bar"
  }
}
```

The `via` header does not show up in the result, while `x-via` does. I would expect both header to be reflected.

sanketsaurav:
If `show_env` is True, then return all the headers without stripping `ENV_HEADERS`. Also adds tests for `/headers` endpoint.

Signed-off-by: Sanket Saurav <sanketsaurav@gmail.com>

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -282,7 +282,7 @@ def view_uuid():
 
 @app.route('/headers')
 def view_headers():
-    """Return the incoming requests's HTTP headers.
+    """Return the incoming request's HTTP headers.
     ---
     tags:
       - Request inspection
@@ -290,10 +290,10 @@ def view_headers():
       - application/json
     responses:
       200:
-        description: The Rrquest's IP Address.
+        description: The request's headers.
     """
 
-    return jsonify(get_dict('headers'))
+    return jsonify(get_headers())
 
 
 @app.route('/user-agent')
--- a/test_httpbin.py
+++ b/test_httpbin.py
@@ -265,6 +265,30 @@ class HttpbinTestCase(unittest.TestCase):
         self.assertEqual(
             response.headers.get('Access-Control-Allow-Headers'), 'X-Test-Header'
         )
+
+    def test_headers(self):
+        headers = {
+            "Accept": "*/*",
+            "Host": "localhost:1234",
+            "User-Agent": "curl/7.54.0",
+            "Via": "bar"
+        }
+        response = self.app.get('/headers', headers=headers)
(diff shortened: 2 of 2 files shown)


test_httpbin.py
```diff
@@ -265,6 +265,29 @@ def test_set_cors_allow_headers(self):
...
+    def test_headers(self):
+        headers = {
+            "Accept": "*/*",
+            "Host": "localhost:1234",
+            "User-Agent": "curl/7.54.0",
+            "Via": "bar"
+        }
+        response = self.app.get('/headers', headers=headers)
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue({'Accept', 'Host', 'User-Agent'}.issubset(set(response.json.keys())))
```
nateprewitt: It may be good to have an explicit check here that `Via` is not included in this set. Otherwise, these two tests could return the same header set and we wouldn't know.

httpbin/core.py
```diff
@@ -293,7 +293,7 @@ def view_headers():
         description: The Rrquest's IP Address.
     """
 
-    return jsonify(get_dict('headers'))
+    return jsonify(get_headers())
```
nateprewitt: @sigmavirus24, it looks like `get_dict('headers')` is already using `get_headers` [under the hood](https://github.com/requests/httpbin/blob/63a649f753e625feac9196115a8ba74fd36864fe/httpbin/helpers.py#L192). It also already supports a `show_env` arg which seems to work on production httpbin right now. Are we sure this change is needed?

test_httpbin.py
sanketsaurav: @nateprewitt Added an explicit check on L279.

httpbin/core.py
```diff
@@ -290,10 +290,10 @@ def view_headers():
       - application/json
     responses:
       200:
-        description: The Rrquest's IP Address.
+        description: The Request's IP Address.
```
nateprewitt: Good catch on the typo. This looks like it's actually a bad description too. We want this to be "The request's headers."

sanketsaurav:
@nateprewitt Fixed the docstring as well! :)

PR #458, opened 2018-05-19
Dockerfile improvements

steffenschroeder:
### Before the fix
Changing any file lead to repeatedly creating new layers for `RUN apt update -y ` and `apt install python3-pip -y` which is slow and takes up much space

### After the fix
Apply best practices from https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run and https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#build-cache

 `RUN apt update -y ` and `apt install python3-pip -y` is cached. The `docker build` becomes faster, fewer layers are created. Users of the Dockerfile and `docker-compose` need to download less from Dockerhub if anything in this project is changed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
 FROM ubuntu:18.04
 
-ADD . /httpbin
-
 RUN apt update -y
 RUN apt install python3-pip -y
+
+ADD . /httpbin
+
 RUN pip3 install --no-cache-dir gunicorn /httpbin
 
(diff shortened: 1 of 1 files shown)


steffenschroeder:
Could be even improve with further PR to run `RUN pip3 install --no-cache-dir gunicorn /httpbin` only if `setup.py` or `Pipfile` or `Pipfile.lock` changed

PR #468, opened 2018-05-25
Pretty-print JSON responses

hkosova:
Previously httpbin used to return pretty-printed JSON, which was easy to read when viewing raw responses. But now the responses are minified.

httpbin uses Flask, and Flask 1.0 [changed the default](http://flask.pocoo.org/docs/1.0/changelog/#version-1-0) for [`JSONIFY_PRETTYPRINT_REGULAR`](http://flask.pocoo.org/docs/1.0/config/#JSONIFY_PRETTYPRINT_REGULAR) to `False`, which is what probably caused the change in httpbin's behavior.

This PR explicitly sets [`JSONIFY_PRETTYPRINT_REGULAR`](http://flask.pocoo.org/docs/1.0/config/#JSONIFY_PRETTYPRINT_REGULAR) to `True` to pretty-print JSON responses.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -55,6 +55,7 @@ tmpl_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
 
 app = Flask(__name__, template_folder=tmpl_dir)
 app.debug = bool(os.environ.get('DEBUG'))
+app.config['JSONIFY_PRETTYPRINT_REGULAR'] = True
 
 app.add_template_global('HTTPBIN_TRACKING' in os.environ, name='tracking_enabled')
 


sigmavirus24:
Bandwidth isn't an issue for our hosting platform, although I'm sure they'd appreciate it if we didn't enable this by default.

I wonder if this could be turned into an environment variable that could be toggled similar to other config vars we have.

Alternatively, I wonder if we could user-agent detection for this (like GitHub's API does)

Lucretiel:
I'll throw out there that, for those who are using this on the CLI (such as with cURL), python has a built-in pipeline pretty printer: `curl ... | python -m json.tool`

kennethreitz:
✨🍰✨

PR #490, opened 2018-07-11
Update README.md

GammaGames:
Add instructions to run locally since the website keeps going down and this makes it easier for users that might not be as familiar with docker


GammaGames:
I realized earlier today that it might be useful to list the endpoints in the readme as well, I might add those and open another pr later.

PR #492, opened 2018-07-25
Revert #457 "Support `show_env` param in `/headers`"

### LINKED ISSUE #487: Different results for /headers between docker and httpbin.org
I'm getting different results for the same request (tested by modify /etc/hosts).

Docker returns:

```
{
	"Abc": "def",
	"Connection": "close",
	"Content-Type": "application/x-www-form-urlencoded",
	"Foo": "bar",
	"Host": "httpbin.org"
}
```

The httpbin.org site returns:

```
{
	"headers": {
		"Abc": "def",
		"Connection": "close",
		"Content-Type": "application/x-www-form-urlencoded",
		"Foo": "bar",
		"Host": "httpbin.org"
	}
}
```

That headers wrapper object being inconsistently present is causing some havoc for tests in our project, where developers use the live site, but CI is using Docker.

javabrett:
Reverts #457. Fixes #487.

The docstring typo fixes are retained, along with the additional tests,
modified to revert to the original `"headers": {` dict format.

The reverted change `get_dict('headers')` -> `get_headers()` was not required. The
query-string `?show_env=1` was already effective in allowing env headers to be
returned.

This partially reverts commit 31ffe79981a079d6605352d2d84de2f2d27ab6db.

That this change was not required (other than the tests, thanks) was hinted-at by @nateprewitt in the review for #457:

> I'm not sure if get_dict change is needed, although it probably won't hurt anything either. I'd definitely like to see the tests merged at the least though.

In fact the `show_env` pass-through option for filtered env headers has been there since that feature was added in bb094d4a .

@sanketsaurav please chime-in here if this analysis seems off.  Your PR might have fixed something for you.  /cc @Fishbowler .

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -342,7 +342,7 @@ def view_headers():
         description: The request's headers.
     """
 
-    return jsonify(get_headers())
+    return jsonify(get_dict('headers'))
 
 
 @app.route("/user-agent")
--- a/test_httpbin.py
+++ b/test_httpbin.py
@@ -279,7 +279,7 @@ class HttpbinTestCase(unittest.TestCase):
         }
         response = self.app.get('/headers', headers=headers)
         self.assertEqual(response.status_code, 200)
-        self.assertTrue({'Accept', 'Host', 'User-Agent'}.issubset(set(response.json.keys())))
+        self.assertTrue({'Accept', 'Host', 'User-Agent'}.issubset(set(response.json['headers'].keys())))
         self.assertNotIn('Via', response.json)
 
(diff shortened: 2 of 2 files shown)


sanketsaurav:
Hi @javabrett,

Yes, your analysis is accurate indeed. I guess we can revert the change to the view logic if we need to retain the existing dict format. I guess we missed that while discussing on my PR the last time around.

Fishbowler:
👍

kennethreitz:
✨🍰✨

kennethreitz:
deployed.

PR #496, opened 2018-07-27
Dockerfile: Resolve deps using pipenv/Pipefile(.lock)

### LINKED ISSUE #493: Docker image build not deterministic/reproducible, uses setup.py not Pipenv.lock, differs from dev env
TL;DR: current Docker image built today will include `greenlet` version `0.14.4`, whereas dev environments get `0.14.3` from `Pipenv.lock`.

This issue:

* Is based on the thinking that the Docker image represents an _Application_ build/deployment of _httpbin_, as opposed to _httpbin_ as a _Library_, ref https://docs.pipenv.org/advanced/#pipfile-vs-setuppy
* Assumes that the Docker build should be good for both production and development uses (but might be better-suited to one or the other)
* Assumes where possible, and within some of the challenges presented by Docker to make reproducible builds (such as moving tags for base images), that the _httpbin_ Docker image build should make best-efforts to be deterministic and reproducible

To illustrate this issue I am making-use of the fact that for current master @ fce4d8339702b5829e8c1dc7b2fff38e6a8b56bb `Pipfile.lock`, [`greenlet` is at version `0.14.3`](https://github.com/requests/httpbin/blob/660801081cc485754720a11cc83a03c4c1df10e0/Pipfile.lock#L164) and a new version `0.14.4` has been released since the last `pipenv lock` was committed.

Part of my motivation for this issue is that I have been learning `pipenv` and want to develop some best-practices.  The install of `httpbin` into a Docker image seems interesting, since it runs under `gunicorn` effectively as a _Library_, but the deployment context is more of an _Application_, and we might expect reproducible builds and installations.

Expectation: When developing using `pipenv`, the versions we get from `Pipenv.lock` we expect to be the same versions in a Dockerfile built from the same Git revision today, and the same Git revision in the future.

Observation: When the current master revision was committed Thu Jul 12 2018, `greenlet` version `0.14.3` was the newest released and is the version in `Pipfile.lock` and would also appear in the Docker image.  When building today, `pipenv sync` and `pipenv shell` produce version `0.14.3`, but the Docker image via `setup.py` and `setuptools` gets the latest version of `greenlet`, `0.14.4`.

Assuming this should be fixed, I have a couple of approaches which work by either generating a `requirements.txt` or running a `pipenv` within the container.  But wanted to get feedback first.

javabrett:
Been experimenting with `pipenv` for reproducible Docker image builds and came-up with this approach for _httpbin_.

It makes a couple of Docker cacheability improvements - mainly because `Pipfile` and `Pipfile.lock` are added on their own before deps are installed, so changes to the rest of the source-code won't cause a cache-miss and re-download of all the dependencies.

I'm curious as to whether this is an intended use of `pipenv` ... it seems to help solve the problems for installing as _Applications_ (versus _Libraries_) with pinned versions for dependencies for a specific build.

With reference to #493, with this change the built Docker image should contain `greenlet` version `0.14.3` (the locked version), and not the latest `0.14.4` which would be resolved by `setup.py`.

Fixed #493.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,18 @@ LABEL version="0.9.2"
 LABEL description="A simple HTTP service."
 LABEL org.kennethreitz.vendor="Kenneth Reitz"
 
-RUN apt update -y && apt install python3-pip -y
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
 
-EXPOSE 80
+RUN apt update -y && apt install python3-pip git -y && pip3 install --no-cache-dir pipenv
+
+ADD Pipfile Pipfile.lock /httpbin/
+WORKDIR /httpbin
+RUN /bin/bash -c "pip3 install --no-cache-dir -r <(pipenv lock -r)"
 
 ADD . /httpbin
+RUN pip3 install --no-cache-dir /httpbin
 
-RUN pip3 install --no-cache-dir gunicorn /httpbin
+EXPOSE 80
 
 CMD ["gunicorn", "-b", "0.0.0.0:80", "httpbin:app", "-k", "gevent"]


Dockerfile
```diff
@@ -5,12 +5,18 @@ LABEL version="0.9.2"
...
+
+ADD Pipfile Pipfile.lock /httpbin/
+WORKDIR /httpbin
+RUN /bin/bash -c "pip3 install --no-cache-dir -r <(pipenv lock -r)"
 
 ADD . /httpbin
+RUN pip3 install --no-cache-dir /httpbin
 
-RUN pip3 install --no-cache-dir gunicorn /httpbin
+EXPOSE 80
```
steffenschroeder: I wonder why the was moved here? (Also the ‘WORKDIR’). Couldn’t we move them again to top so we can avoid adding additional  docker layers for these if ‘Pipfile.lock’ changes?

PR #503, opened 2018-08-15
Added basic-auth path to allow testing with empty password. Fixed #502.

### LINKED ISSUE #502: Basic auth without both username and password
https://httpbin.org/basic-auth/foo/bar works for testing basic auth with a username and a password.

However, all of the following give 404 not found errors:

* https://httpbin.org/basic-auth/foo
* https://httpbin.org/basic-auth/foo/
* https://httpbin.org/basic-auth/foo//
* https://httpbin.org/basic-auth//bar

Is there a way to test these features of basic auth with httpbin?

javabrett:
Fixed #502.


nateprewitt:
Hey @javabrett, would you mind adding some testing to demonstrate exactly what we expect to see here. I'm assuming no username or password would just result in `Authorization: Basic` with no checksum?

javabrett:
The existing coverage is in `test_set_cors_credentials_headers_after_auth_request`, which is where I'd added some tests already - this method is named specifically for CORS testing, but I'm co-opting it.

The existing test didn't really check an authentication following the challenge, so I've added a commit which checks that, for all cases of `foo:bar`, `foo:` and `:`.  You'll note that the correct auth response for an empty username/password is `Basic Og==` (encoded colon).  Headers can be checked here: https://www.blitter.se/utils/basic-authentication-header-generator/ .

javabrett:
Any desire to review/merge ... else closing?

PR #505, opened 2018-08-16
Implement GSSAPI authentication endpoint

mkomitee:
Allow GSSAPI/Kerberos authentication testing.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -33,6 +33,12 @@ from werkzeug.wrappers import BaseResponse
 from werkzeug.http import parse_authorization_header
 from flasgger import Swagger, NO_SANITIZER
 
+try:
+    import gssapi
+    HAS_GSSAPI = True
+except ImportError:
+    HAS_GSSAPI = False
+
 from . import filters
 from .helpers import (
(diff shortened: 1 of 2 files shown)


yschimke:
Are there instructions for running a test client against this endpoint?  Seems like one of the harder ones to configure, e.g. for basic-auth you know to supply the credentials in the path and then supply those with Basic auth encoded header.

Is it as simple for clients here?

javabrett:
@mkomitee is it possible to write a test for this new endpoint?

mkomitee:
It is  possible, but it's a bit involved. It requires a functioning kdc. Someone set something up for testing purposes in the https://github.com/requests/requests-kerberos project with https://github.com/requests/requests-kerberos/blob/master/.travis.sh.

mkomitee:
It is  possible, but it's a bit involved. It requires a functioning kdc. Someone set something up for testing purposes in the https://github.com/requests/requests-kerberos project with https://github.com/requests/requests-kerberos/commit/29dcf74d072eed6df74abe216fbfcfe45de29758, but it involves using a docker container.

PR #507, opened 2018-08-31
Fix a ResourceWarning: unclosed file in core.py

BoboTiG:
Hello :)

A little fix for an unclosed file resource warning.


sigmavirus24:
Thanks @BoboTiG !

PR #512, opened 2018-09-21
remove duplicated "X-Forwarded-For" from ENV_HEADERS

georgexsh:


georgexsh:
@nateprewitt alright, but IMHO, remove those so-called "ENV_HEADERS" is a terrible idea from the beginning.

georgexsh:
@nateprewitt thanks, I have noticed that this behavior exists from the beginning, nevertheless, it should be documented.

PR #516, opened 2018-09-29
Add SHA1 digest auth support

jameshilliard:
Requests supports this algo [here](https://github.com/requests/requests/blob/943a5c8e89db1758ae24adbbedacb3b05c32df4a/requests/auth.py#L150).

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -1075,7 +1075,7 @@ def digest_auth_nostale(qop=None, user="user", passwd="passwd", algorithm="MD5")
       - in: path
         name: algorithm
         type: string
-        description: MD5, SHA-256, SHA-512
+        description: MD5, SHA, SHA-256, SHA-512
         default: MD5
     produces:
       - application/json
@@ -1111,7 +1111,7 @@ def digest_auth(
       - in: path
         name: algorithm
         type: string
-        description: MD5, SHA-256, SHA-512
+        description: MD5, SHA, SHA-256, SHA-512
         default: MD5
       - in: path
         name: stale_after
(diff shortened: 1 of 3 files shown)


nateprewitt:
Hi @jameshilliard, thanks for pointing this bit of code out. I'm actually not sure why that section exists in Requests. The commit is quite old (>6 years) and as far as I can tell from RFC [2617](https://tools.ietf.org/html/rfc2617), [7616](https://tools.ietf.org/html/rfc7616), and a google search, Digest Auth has never supported SHA-1.

Given that SHA-1 is now cryptographically unsafe, it's probably not a good idea to promote testing it's usage. It looks like we have full coverage according to RFC [7616 3.2](https://tools.ietf.org/html/rfc7616#section-3.2), so I'm going to close this out. Thanks again though!

jameshilliard:
@nateprewitt SHA-1 digest auth is supported by twisted(with MD5 being the only other supported option which is much less secure) when used as a http server so it's probably still useful to have this for automated client regression testing(which is why I was adding this).

jameshilliard:
@nateprewitt 
> just to be clear, SHA-1 is no safer than MD5 since the revelation of Shattered

I'm fairly certain this isn't true at all, shattered means SHA-1 is insecure for a number of use cases but due to the protocol construction of http digest auth I don't think a hash collision vulnerability like shattered is actually a problem here since the attacker doesn't have plain text access which appears to be required for those types of attacks. Even MD5 is not fully broken when it comes to digest authentication from my understanding(there are way more known attacks against MD5 than SHA-1).

javabrett:
I don't have a burning desire to add SHA-1 for Digest auth, but it is worth noting that [RFC 7616](https://tools.ietf.org/html/rfc7616) does establish an [IANA Hypertext Transfer Protocol (HTTP) Digest Algorithm Values
](https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml), which contains SHA-1 as an algorithm.  The RFC only directly mentions the initial and commonly-used/default registry entries.

PR #517, opened 2018-10-22
Dockerized tests

avishayp:
In this PR:

 * Move docker base image to `python3-alpine` instead of ubuntu.
 * Reduce docker context size with more aggressive `.dockerignore`
 * Add a service that runs `test_httpbin.py` to `docker-compose.yml`

### why
I was not able to run `tox` tests on my mac. I did not dig into that - docker provides more reproducible  environment for running unittests.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,5 @@
-Dockerfile
-.git
+**
+
+!/httpbin
+!/requirements.txt
+!/test_httpbin.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,23 @@
-FROM ubuntu:18.04
+FROM python:3-alpine
 
 LABEL name="httpbin"
 LABEL version="0.9.2"
 LABEL description="A simple HTTP service."
 LABEL org.kennethreitz.vendor="Kenneth Reitz"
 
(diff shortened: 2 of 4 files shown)


javabrett:
Disclaimer: I'm not a collaborator on this project and don't have any say on what gets merged.

Consider decomposing this PR - the main thrust, which is to run tests, should be independent of the Docker base/build changes.

The tests should be run as part of the Docker build, not as a Compose service, e.g.:

```
diff --git a/Dockerfile b/Dockerfile
index 6e391a6..6b4da16 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ EXPOSE 80
 
 ADD . /httpbin
 
-RUN pip3 install --no-cache-dir gunicorn /httpbin
+RUN pip3 install --no-cache-dir gunicorn /httpbin && \
+    python3 /httpbin/test_httpbin.py
```

... or perhaps better, arrange for the equivalent of `python3.6 setup.py test` as mentioned in #448.

kennethreitz:
let's break this out, keep the image the same for now

avishayp:
Replaced with https://github.com/requests/httpbin/pull/520 as requested.

PR #519, opened 2018-10-27
Optimize Dockerfile for reducing the size and drop root

jeffrey4l:
- move base from ubuntu to python:3.6.7-slim which decrease the image
  size from 553MB to 192 MB
- Use 8080 for the gunicorn listen port, which could make the image run
  without root priviledge. This is import if you wanna run httpbin in
  OpenShift.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,14 @@
-FROM ubuntu:18.04
+FROM python:3.6.7-slim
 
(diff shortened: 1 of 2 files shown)


jeffrey4l:
How about the drop-root part?

kennethreitz:
I think running it on port 80 is professional

PR #520, opened 2018-10-28
Run unittests in docker container

avishayp:
### why
I was not able to run `tox` tests on mac. I assume this is fixable, but running tests in a container gives a more predictable environment.

### what

 * Added `test` service to `docker-compose.yml` that runs the unittests.
 * Added note in `README.md`

PR #521, opened 2018-10-28
Run tests during Docker image build

javabrett:
Inspired by PRs from @avishayp #520 and #517.

This change just runs the unit tests in the Docker container during the image build, immediately following the install.  This is good because:

- For developers doing Docker-only builds and not running the tox/test toolchain, it prevents accidentally creating an image which has failing tests.
- Prevents undetected failures in the Docker Hub automated image build.
- Catches any Docker-specific issues or failures.
- These tests are cheap to run, we should run them as part of the Docker image build as well as other builds.

I added an optional build arg `IGNORE_TEST_FAILURES=false`, so you can `docker build -t httpbin --build-arg IGNORE_TEST_FAILURES=true .` if you need to force an image build with failing tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,9 @@ RUN /bin/bash -c "pip3 install --no-cache-dir -r <(pipenv lock -r)"
 ADD . /httpbin
 RUN pip3 install --no-cache-dir /httpbin
 
+ARG IGNORE_TEST_FAILURES=false
+RUN cd /httpbin;python3 test_httpbin.py || $IGNORE_TEST_FAILURES
+
 EXPOSE 80
 
 CMD ["gunicorn", "-b", "0.0.0.0:80", "httpbin:app", "-k", "gevent"]


avishayp:
Separating `RUN` commands would make better use of image layers.

```bash
RUN pip3 install --no-cache-dir gunicorn
RUN pip3 install --no-cache-dir /httpbin
RUN cd /httpbin;python3 test_httpbin.py || $IGNORE_TEST_FAILURES
```

That way changing `IGNORE_TEST_FAILURES` only invalidates the tests layer.

javabrett:
> That way changing IGNORE_TEST_FAILURES only invalidates the tests layer.

My assumptions here are:
- `IGNORE_TEST_FAILURES=true` will only be set extremely rarely in the case where you absolutely need to force an image build when there is a failing test, e.g. maybe to debug that test.  I don't expect it to change frequently.
- If `IGNORE_TEST_FAILURES` is changing, presumably there are source and test-source changes afoot, which will invalidate the pre-test cache anyway.

It is true that there are a number of build-efficiencies that could be made in various ways regarding the build cache, but I'm leaving them separate from this PR, which shouldn't make any of those worse.  I have another PR related to cache and some other things but they have not been worked-on or reviewed for a while.

javabrett:
Rebased this to latest `Dockerfile` changes.  Ready for review.  I think it's worth running the tests during the Docker build.

das7pad:
> For developers doing Docker-only builds and not running the tox/test toolchain, it prevents accidentally creating an image which has failing tests.

Simply do not merge broken code into the default branch?

> Prevents undetected failures in the Docker Hub automated image build.

Running a test suite on a new image build before pushing to Docker Hub would fit better.

> Catches any Docker-specific issues or failures.

Any examples?

avishayp:
---
It's better to be roughly right than precisely wrong

javabrett:
@das7pad I'm not sure whether you are voting against this change, but I suspect-so.  Can you offer any additional details/objections or suggested-changed?  Thanks.

> Simply do not merge broken code into the default branch?

Yes it would be nice if bugs never got written or merged, but that's why I want the tests to also run in the Docker build, since that's another place to catch them.

> Running a test suite on a new image build before pushing to Docker Hub would fit better.

How should we do that better than running the test in the image during the image build?  With this change, the existing Docker Hub automated build should do exactly this - if the tests fail, a new image won't be pushed, and a build failure notification should occur.  To otherwise incorporate this would probably require an external orchestration e.g. Jenkins, and I don't see how that should be necessary.

> Any examples?

#403 comes to mind.

@avishayp 

> It's better to be roughly right than precisely wrong

Which one is this PR? :).

Anyone else - would be good to get reviews, +ve or -ve for this.

das7pad:
@javabrett Yes, I am voting against this change.

Instead of running the tests during the build of the application image, we could run the tests in the new image.
Travis could do this for us: build the image once and run the suite afterwards.

This way a build would fail for the PR which introduced the bug. With the change of this PR a build would fail for a random commit that eventually triggered the Docker Hub build.

But I am still not sure which type of _bug_ a dockerised test should detect. Travis already runs in a sandbox.

#403 is not related to docker.

PR #525, opened 2018-11-12
remove http://requestb.in from README.md

nosoyyo:
http://requestb.in seems nothing there for now.


cabello:
Makes sense, it's currently offline.

AaronRobson (approved):

gaul:
See #557.

PR #543, opened 2019-03-09
Adds server network info app.route and to /anything

simonkowallik:
Adds server network information (hostname, ip addresses and ip aliases) to a new app.route (`/serverinfo`) as well as to the existing `/anything` route.
The new route did not match any existing swagger tag, hence created a new tag.

Having the server network information available is very helpful in some networking tests, especially when servers need to uniquely identified, for example in load balancing scenarios.

An example request to /serverinfo produces the following response:
```
{
  "aliaslist": [],
  "hostname": "8fee1b6a3fa5",
  "ipaddrlist": [
    "172.17.0.2"
  ]
}
```

However Hostname, IP addresses + aliases can be considered sensitive information, hence this is a possible drawback and security implication.
I considered making this optional (e.g. based on environment variable) or entirely replace it with a unique 'instance' ID but was first looking for opinions/feedback.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -48,6 +48,7 @@ from .helpers import (
     parse_multi_value_header,
     next_stale_after_value,
     digest_challenge_response,
+    get_serverinfo,
 )
 from .utils import weighted_choice
 from .structures import CaseInsensitiveDict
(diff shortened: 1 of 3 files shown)


simonkowallik:
closing in favour of #565 due to the concerns listed in the initial comment

PR #548, opened 2019-04-16
Select status code with anything

cabello:
Hi @kenneth-reitz 👋 

Thanks for this awesome project and the lovely `requests` library.

I would like to tweak a bit the behaviour of `/status/<codes>` routes by adding new functionality without breaking changes.

Use case: test what would happen in a service oriented architecture if a service only returned a set of status codes for any incoming request, for example "504 Gateway Timeout".

With this change users are able to configure the client base endpoint to be `https://httpbin.org/status/504` and if the regular client appends something like `/api/v1/users/123` they would get `504` back as opposed to `404 Not Found` at the moment.

Thanks for considering my contribution 🙇

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -729,7 +729,10 @@ def stream_n_messages(n):
 @app.route(
     "/status/<codes>", methods=["GET", "POST", "PUT", "DELETE", "PATCH", "TRACE"]
 )
-def view_status_code(codes):
+@app.route(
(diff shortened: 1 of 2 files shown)


cabello:
I saw the build was failing locally for me on these two tests that are unrelated to my change, should I attempt to fix it? It looks like `Content-Length` is not present in the response.

simonkowallik:
> I saw the build was failing locally for me on these two tests that are unrelated to my change, should I attempt to fix it? It looks like `Content-Length` is not present in the response.

I had the same issue and created issue #554 and pull request #555.

PR #553, opened 2019-05-13
Remove raven from setup.py

yan12125:
Usage of raven is removed in https://github.com/postmanlabs/httpbin/commit/e5433806757a09d58e48935c081db80170296ab5


yan12125:
Seems this project has been dead. I will no longer keep an eye on this. Anyone can use up my patch in any place if they wish.

PR #555, opened 2019-05-18
Fix tests for werkzeug>=0.15.1 and update Pipfile and Pipfile.lock

### LINKED ISSUE #554: tests broken: content-length no longer present with werkzeug >= 0.15.1
Recent travis-ci tests fail. In particular `test_get` and `test_anything`.
Both tests expect the `Content-Length` header to be present:
https://github.com/postmanlabs/httpbin/blob/f8ec666b4d1b654e4ff6aedd356f510dcac09f83/test_httpbin.py#L151 https://github.com/postmanlabs/httpbin/blob/f8ec666b4d1b654e4ff6aedd356f510dcac09f83/test_httpbin.py#L165

It seems there is a behaviour change in werkzeug between version 0.14.1 and 0.15.1 (and newer), where Content-Length is only set when non zero.

Pipfile requires werkzeug >= 0.14.1:
https://github.com/postmanlabs/httpbin/blob/f8ec666b4d1b654e4ff6aedd356f510dcac09f83/Pipfile#L12

I first noticed this for my pull requests, where #543 passed the existing tests just fine. This test ran with werkzeug 0.14.1: https://travis-ci.org/postmanlabs/httpbin/jobs/504080576#L205

Where the following test failed and used werkzeug 0.15.1:
https://travis-ci.org/postmanlabs/httpbin/jobs/510702985#L205

Either locking werkzeug to 0.14.1 or updating the tests and requiring werkzeug >= 0.15.1 would probably work fine also the later is preferable.

simonkowallik:
address issue #554
- update test_httpbin.py to reflect new behaviour of werkzeug >=0.15.1
- require werkzeug>=0.15.1 in Pipfile
- update Pipfile.lock with werkzeug 0.15.1 hashes


gaul (commented):
Should this also upgrade the dependency in `setup.py`?

gaul (approved):

simonkowallik:
> Should this also upgrade the dependency in `setup.py`?

It definitely should and now it does! Thanks!

sgaynetdinov (approved):

AaronRobson (approved):

AaronRobson (approved):

sgaynetdinov (approved):

AaronRobson:
Are there any maintainers about that would kindly merge this for us?

AaronRobson (approved):

AaronRobson (approved):

sgaynetdinov (approved):

PR #556, opened 2019-05-27
Use Brotli instead of brotlipy

felixonmars:
brotlipy is stuck at brotli 0.6 and upstream is inactive. Let's switch
to the official binding which is up-to-date and has same interfaces.


felixonmars:
The CI failure should be https://github.com/postmanlabs/httpbin/issues/554

z3ntu:
bump

PR #557, opened 2019-05-29
Changing requestb.in -> requestbin.com

dylburger:
requestb.in is defunct, but requestbin.com offers a comparable bin service.


gMan1990:
see https://docs.postman-echo.com/?version=latest

PR #573, opened 2019-08-22
Replace localhost with constant

gaul:
This allows overriding a single constant when using an external
httpbin.  References gaul/java-httpbin#4.


gaul:
CI fails due to #554.

AaronRobson (approved):

PR #588, opened 2019-12-04
Helm chart, and use alpine image.

JanRK:
Added a helm chart for easy deploy in Kubernetes.

Changed Dockerfile to use alpine image, and made it multistage.
Also changed dockerfile to be able to run without root and on a readonly filesystem.

Image size:
Multistage  : 63.3MB
Alpine : 68.5MB
Normal : 238.2MB


helmchart/values.yaml
```diff
@@ -0,0 +1,71 @@
+# Default values for httpbin.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: gcr.io/mvno-e3a61b6e-3/httpbin
```
mhamann: This should be a pointer to Docker Hub, I think.

mhamann:
This looks really good, @JanRK. Thanks for doing this work!

Ideally, we could get this into the "official" helm repo at some point, but this'll definitely work for now.

+1 for use of alpine.

helmchart/values.yaml
JanRK: Ah yes. Since the chart has readonly fs and no root, I couldn't point to the official image. So had it pointing to my own non public repo. Changed it to a public docker hub build.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM python:alpine3.10 AS compile-image
 
 LABEL name="httpbin"
 LABEL version="0.9.2"
```
bergerx: The labels should go into the resulting container rather than the compile container.

Dockerfile
```diff
@@ -8,15 +8,26 @@ LABEL org.kennethreitz.vendor="Kenneth Reitz"
...
 RUN pip3 install --no-cache-dir /httpbin
 
-EXPOSE 80
 
-CMD ["gunicorn", "-b", "0.0.0.0:80", "httpbin:app", "-k", "gevent"]
+FROM python:alpine3.10
+COPY --from=compile-image /opt/venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+
+EXPOSE 8080
```
bergerx, line 31: Changing the default port to 8080 from 80 is actually a breaking change for the resulting container. If container version is tightly coupled with httpbin's version number, it could worth bumping the right version field. I don't know what release convention httpbin uses and how its released, just my 2 cents.

bergerx (changes requested):
Would you mind moving/renaming the chart directory as this: `helmchart` --> `charts/httpbin`.
As per https://helm.sh/docs/topics/chart_best_practices/conventions/#chart-names :
> The directory that contains a chart MUST have the same name as the chart. Thus, the chart nginx-lego MUST be created in a directory called nginx-lego/. This is not merely a stylistic detail, but a requirement of the Helm Chart format.

Dockerfile
JanRK: Thanks, fixed.

JanRK:
> Would you mind moving/renaming the chart directory as this: `helmchart` --> `charts/httpbin`.
> As per https://helm.sh/docs/topics/chart_best_practices/conventions/#chart-names :
> 
> > The directory that contains a chart MUST have the same name as the chart. Thus, the chart nginx-lego MUST be created in a directory called nginx-lego/. This is not merely a stylistic detail, but a requirement of the Helm Chart format.

Fixed

Dockerfile
JanRK: Yeah, the change was made so gunicorn could run without root. 

I guess there are two solutions:
Make a 2nd docker image for the helm chart, and leave the original for docker users
Change it back to running as root.

bergerx: I'd say this is the maintainers' decision.
I just provided my opinion, if they are fine with the change, it's ok.

bergerx:
I'm seeing that the travis checks has failed.
Did you check if its this PR breaking them or were they broken before?

23ewrdtf:
Hi,
I'm in need of a helm chart for httpbin so happy to test if you have it.

Adam

JanRK:
The chart is here
https://github.com/JanRK/httpbin/tree/master/charts/httpbin

bergerx (approved):
LGTM

bergerx:
@mhamann do you mind taking another look into this one?

charts/httpbin/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: httpbin
+description: A Helm chart for Kubernetes
```
mhamann, line 3: Should change this to something like "Official Helm chart for the httpbin service"

mhamann (approved):
lgtm

pklaus:
What a pity the project is inactive at the moment.

I like the updates on the Dockerfile (don't care too much about Helm, although I also intend to deploy to Kubernetes).
I cloned the branch from this pull request (HEAD currently at commit 5b0a7bb) and built a multi-arch image using `docker buildx build -t pklaus/mods:httpbin --platform "linux/amd64,linux/386,linux/arm64,linux/arm/v7" --pull --push .`

The Docker Image can be found [on the Docker Hub](https://hub.docker.com/r/pklaus/mods/tags?page=1&name=httpbin) if anyone wants to save the step of building it from scratch.

PR #628, opened 2020-11-25
Add factorial endpoint for CPU stresstest

cristiklein:
Hi,

We are using httpbin as a backend to stress-test reverse proxies, service meshes, circuit breakers, etc. It would be very useful for us if httpbin could be instructed to produce some "real" load. Would you be open to adding a factorial endpoint?

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -10,6 +10,7 @@ This module provides the core HttpBin experience.
 import base64
 import json
 import os
+import math
 import random
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


gaul:
This will be a DDoS vector for the public-facing service.

cristiklein:
I wasn't aware that this project is supposed to be public-facing. My PR is definitely not something you want public-facing.

Would you accept a PR in which the factorial endpoint is enabled by an environment variable (default disabled)?

gaul:
httpbin.org hosts this software:

```
$ curl -I http://httpbin.org/status/418
HTTP/1.1 418 I'M A TEAPOT
```

PR #629, opened 2020-12-02
Replace brotlipy with brotlicffi

mgorny:
The brotlipy package has been renamed to brotlicffi.  Update the imports
and dependencies accordingly.  The major advanage of the new package
is that it no longer collides with the Python bindings provided
by brotli itself.


mgorny:
Note: if you want to continue supporting Python 2.7, this requires https://github.com/python-hyper/brotlicffi/pull/172 to be merged and released first.

Issue #577, opened 2019-09-04
Content encoding of POST request body is ignored


pimterry:
If a request a POSTed with a gzipped body, and a `content-encoding: gzip` header, the content is not decoded.

For example when posting to `/post` with an `content-type: application/json` gzipped body & content encoding, the response contains `"data": "data:application/octet-stream;base64,BASE64GZIPPEDDATA",` and `"json": null`.

If there is no encoding used, `json` shows the parsed json correctly.

raphaeleidus:
web servers should not decompress gzipped content they dont trust. Webservers that do can get gzip bombed.
that is a crafted payload that is extremely duplicative for example a 20GB file filled with just the letter `a` over and over. When gzipped it shrinks to a really small size, if transferred to the server, the server has no idea how big the uncompressed content will be, it might arrive with only a few hundred kb, but then when it decompresses it immediately takes up all the memory and crashes. 
for httpbin.org it should not expand compressed request bodies, but I could see a config flag for the docker version that allows a private instance to do this.

more info: https://en.wikipedia.org/wiki/Zip_bomb

raphaeleidus:
@pimterry I submitted a PR with support for this but it will only be for local usage not for the public facing instances due to attack vulnerability

PR #648, opened 2021-07-08
feat: Add optional feature to decompress encoded request body

raphaeleidus:
This PR addresses #577 
Because of [Zip Bomb](https://en.wikipedia.org/wiki/Zip_bomb) attacks this feature is disabled by default and should only be turned on when running httpbin locally as follows

`docker run --env UNSAFE_BODY_DECOMPRESSION=1 -p 80:80 kennethreitz/httpbin`

There are valid use cases where a server is going to receive payloads only from a trusted source that will be compressed and this can help testing around those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ docker pull kennethreitz/httpbin
 docker run -p 80:80 kennethreitz/httpbin
 ```
 
+### Optional Env Flags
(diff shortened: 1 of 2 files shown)


Starmania:
Good idea

Issue #647, opened 2021-06-23
Tests are failing with the latest werkzeug 2.0.1


frenzymadness:
```
/usr/bin/python3 test_httpbin.py
....................................FFFFF......................
======================================================================
FAIL: test_redirect_n_equals_to_1 (__main__.HttpbinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/httpbin-0.7.0/test_httpbin.py", line 549, in test_redirect_n_equals_to_1
    self.assertEqual(
AssertionError: 'http://localhost/get' != '/get'
- http://localhost/get
+ /get


======================================================================
FAIL: test_redirect_n_higher_than_1 (__main__.HttpbinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/httpbin-0.7.0/test_httpbin.py", line 527, in test_redirect_n_higher_than_1
    self.assertEqual(
AssertionError: 'http://localhost/relative-redirect/4' != '/relative-redirect/4'
- http://localhost/relative-redirect/4
+ /relative-redirect/4


======================================================================
FAIL: test_redirect_to_post (__main__.HttpbinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/httpbin-0.7.0/test_httpbin.py", line 536, in test_redirect_to_post
    self.assertEqual(
AssertionError: 'http://localhost/post' != '/post'
- http://localhost/post
+ /post


======================================================================
FAIL: test_relative_redirect_n_equals_to_1 (__main__.HttpbinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/httpbin-0.7.0/test_httpbin.py", line 555, in test_relative_redirect_n_equals_to_1
    self.assertEqual(
AssertionError: 'http://localhost/get' != '/get'
- http://localhost/get
+ /get


======================================================================
FAIL: test_relative_redirect_n_higher_than_1 (__main__.HttpbinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/httpbin-0.7.0/test_httpbin.py", line 562, in test_relative_redirect_n_higher_than_1
    self.assertEqual(
AssertionError: 'http://localhost/relative-redirect/6' != '/relative-redirect/6'
- http://localhost/relative-redirect/6
+ /relative-redirect/6


[...]

PR #649, opened 2021-07-19
Fix disabling of location header autocorrect for werkzeug 2+

AdamWill:
In werkzeug 2.0.0 and later, the Location header autocorrection
moved from BaseResponse to Response, so we need to set
`autocorrect_location_header = False` in `Response` not
`BaseResponse`. For now let's just set it in both to be safe,
this doesn't cause any errors at least with 1.0.1 and 2.0.1.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -77,7 +77,10 @@ def jsonify(*args, **kwargs):
 
 
 # Prevent WSGI from correcting the casing of the Location header
+# and forcing it to be absolute. This moved from BaseResponse to
+# Response in werkzeug 2.0.0, so we set both to be safe.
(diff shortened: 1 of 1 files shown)


AdamWill:
Fixes #647

MikeRalphson (approved):

MikeRalphson:
@kinlane This PR is a pre-requisite for getting the tests running again when we move them to GitHub Actions - see https://github.com/MikeRalphson/httpbin/actions/runs/1174257560

AdamWill:
Closing in favor of #674 , as setting both isn't valid any more (plus this wasn't doing it entirely right).

PR #684, opened 2022-09-26
fix Werkzeug 2.1.x compatibility

dshein-alt:
- fix httpbin/core.py: use Responce class instead of BaseResponse
see: https://github.com/pallets/werkzeug/pull/2276

- fix tests: TestClient doesn't provide 'Content-Length' header by default anymore
see: https://github.com/pallets/werkzeug/issues/2347


kloczek:
werkzeug 3.0.0 has been released.

PR #698, opened 2023-03-24
Refactor to alpine

pflaeging:
# httpbin with compact alpine container

The goal is:

- compact secure container based on alpine linux
- runnable in kubernetes / OpenShift / OKD in a secure and scalable fashion
- multi architecture container (at least linux/amd64 and linux/arm64: means modern Intel, modern raspis and modern Macs with Mx processor)

## build

We are building the containers with podman (podman-desktop under MacOS or Windows):

```shell
podman build --manifest quay.io/pflaeging/httpbin:0.9.2-alpine --rm --no-cache --platform linux/amd64 --platform linux/arm64 -f Dockerfile.alpine .
podman manifest push quay.io/pflaeging/httpbin:0.9.2-alpine docker://quay.io/pflaeging/httpbin:0.9.2-alpine --rm
```

 (please replace the `quay.io/pflaeging` part with your own registry place)

## Kubernetes rollout

The application is fully compatible with Kubernetes / OpenShift4 / OKD4.

Example objects are in the folder [./kubernetes/](./kubernetes/).

### DIFF
--- a/Alpine-Readme.md
+++ b/Alpine-Readme.md
@@ -1,0 +1,24 @@
+# httpbin with compact alpine container
+
+The goal is:
+
+- compact secure container based on alpine linux
+- runnable in kubernetes / OpenShift / OKD in a secure and scalable fashion
+- multi architecture container (at least linux/amd64 and linux/arm64: means modern Intel, modern raspis and modern Macs with Mx processor)
+
+## build
+
+We are building the containers with podman (podman-desktop under MacOS or Windows):
+
+```shell
+podman build --manifest quay.io/pflaeging/httpbin:0.9.2-alpine --rm --no-cache --platform linux/amd64 --platform linux/arm64 -f Dockerfile.alpine .
+podman manifest push quay.io/pflaeging/httpbin:0.9.2-alpine docker://quay.io/pflaeging/httpbin:0.9.2-alpine --rm
+```
+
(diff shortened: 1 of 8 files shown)


MikeRalphson:
Could you address the performance and stability concerns raised in https://pythonspeed.com/articles/alpine-docker-python/ ?

We might also need to break up this PR so the unrelated bug / dependency changes can be reviewed separately.

pflaeging:
> Could you address the performance and stability concerns raised in https://pythonspeed.com/articles/alpine-docker-python/ ?

We're using alpine normally not because of built-time or size. It's more a matter of security. Normally alpine images are containing fewer packages and so there less affected to CVE's. But I will also make a config with ubi8 / microubi for the deployment on secure OpenShift installations. We also normally using build container configs. You get a smaller and more controllable image in production.

> 
> We might also need to break up this PR so the unrelated bug / dependency changes can be reviewed separately.

I've addressed 3 tasks in my PR:

- k8s deployment
- newer versions of libs
- alpine Dockerfile with build containers

At least the last 2 are a little dependent, 'cause the build doesn't work with the old versions.
The k8s deployment is more independent.

:peter

retornam:
@MikeRalphson  thanks to [PEP-0656](https://peps.python.org/pep-0656/) and this  [PR](https://github.com/pypa/cibuildwheel/pull/768) that  blog post is no longer true.

PR #700, opened 2023-04-19
Allow overriding the default CORS headers

bartbroere:
By default the `Access-Control-*` response headers are very permissive. For testing this is nice, but I came across the use case to test with more restrictive headers.

I tried to do `GET /response-headers?Access-Control-Allow-Origin=http://example.com`, but it replied with `Access-Control-Allow-Origin: *`. After this PR these headers can be overridden.

### DIFF
--- a/httpbin/core.py
+++ b/httpbin/core.py
@@ -215,20 +215,24 @@ def before_request():
 
 @app.after_request
 def set_cors_headers(response):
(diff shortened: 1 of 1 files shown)


bartbroere:
Even though this fixes my use case, by running `tox` I found out it breaks another assumption (and therefore a few tests):

- [ ] fix responses with multiple headers

I know how to fix it, but I'm not sure if it's worthwhile to do it. The last commit on master dates back to 2018. How maintained is this project?

PR #710, opened 2023-09-27
fix: OpenAPI spec invalid

jspdown:
This PR fixes the generated OpenAPI spec document to be v2.0 compliant.  

The most important changes are the removal of the deprecated/non-standard SmartAPI fields `responsibleOrganization` and `responsibleDeveloper`. And the removal of the non-standard `protocol` field at the top level.

These fields alone prevented the use of HTTPBin spec on most OAS compliant tooling.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,3 +23,4 @@ Patches and Suggestions
 - Dave Challis (https://github.com/davechallis)
 - Florian Bruhin (https://github.com/The-Compiler)
 - Brett Randall (https://github.com/javabrett)
+- Harold Ozouf (https://github.com/jspdown)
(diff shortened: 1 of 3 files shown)


exhuma:
Are all those "required" additions really "required"? Note that I've only quickly squinted at this MR through the prism of the "diff" so I am missing some context. The `int`/`number` change looks definitely correct though.

nb: You might want to move this PR over to https://github.com/psf/httpbin

jspdown:
@exhuma Thanks for your feedback.

The Parameter object defined in the OAS v2 specification has a special requirement when the type of a parameter is `path`:

> If the parameter is [in](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#parameterIn) "path", this property is required and its value MUST be true.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#fixed-fields-7

About [psf/httpbin](https://github.com/psf/httpbin), I wasn't aware of this fork. I will make sure to submit the PR there as well.

immanuelfodor:
Checking the PRs there, could you please send it to the fork, too?

